PR #3, opened 2020-07-27
Unit test capabilities

andrei-marinica:
minor refactoring, testing util

### DIFF
--- a/elrond-wasm/src/io/finish.rs
+++ b/elrond-wasm/src/io/finish.rs
@@ -51,41 +51,45 @@ where
     }
 }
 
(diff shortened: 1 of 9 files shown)


elrond-wasm/src/io/finish.rs
```diff
@@ -115,6 +119,8 @@ impl<T> SCResult<T> {
             SCResult::Err(_) => panic!("called `SCResult::unwrap()`"),
         }
     }
+
```
sasurobert, line 122: delete empty lines

PR #10, opened 2020-09-14
Use newAddresses in adder test

hjorthjort:


examples/adder/mandos/adder.scen.json
```diff
@@ -11,7 +11,14 @@
                     "storage": {},
                     "code": ""
                 }
-            }
+            },
+            "newAddresses": [
+                {
+                    "creatorAddress": "0xacc1000000000000000000000000000000000000000000000000000000001234",
+                    "creatorNonce": "5",
+                    "newAddress": "``adder_address"
```
andrei-marinica: use "address:adder_address" or "address:adder_sc". Addresses need to be 32bytes long, this syntax will make sure the right length is generated

andrei-marinica (dismissed):

andrei-marinica (approved):

PR #23, opened 2020-10-09
Optional owner in setState and getState in mandos tests is now possible

dorin-iancu:


elrond-wasm-debug/tests/mandos/external_steps/external_step_2.step.json
```diff
@@ -3,6 +3,7 @@
     "steps": [
         {
             "step": "setState",
+            "owner": "address:owner_address",
```
andrei-marinica, line 6: Th "owner" belongs to the accounts, not globally.

mandos/src/scenario.rs
```diff
@@ -27,6 +27,7 @@ pub enum Step {
     },
     SetState {
         comment: Option<String>,
+        owner: Option<AddressValue>,
```
andrei-marinica, line 30: Add the owner to the Account/AccountRaw, not here.

andrei-marinica (changes requested):

PR #32, opened 2020-10-26
Crowdfunding template adjustments

camilbancioiu:
Add the `currentFunds` view and the empty file `testnet.toml`.

### DIFF
--- a/examples/crowdfunding/src/lib.rs
+++ b/examples/crowdfunding/src/lib.rs
@@ -75,6 +75,11 @@ pub trait Crowdfunding {
         }
     }
 
+    #[view(currentFunds)]
+    fn current_funds(&self) -> SCResult<BigUint> {
+        Ok(self.get_sc_balance())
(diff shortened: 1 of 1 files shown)


examples/crowdfunding/src/lib.rs
```diff
@@ -75,6 +75,11 @@ pub trait Crowdfunding {
         }
     }
 
+    #[view(currentFunds)]
+    fn current_funds(&self) -> SCResult<BigUint> {
```
sasurobert, line 79: there is special api for getBalance :). no need for such view func

sasurobert (approved):

andreibancioiu (approved):

examples/crowdfunding/src/lib.rs
camilbancioiu: Forgot about that. But it's a good example to have it anyway.

PR #44, opened 2020-11-09
Mandos esdt

dorin-iancu:
ESDT support added to mandos tests + crowdfunding contract esdt example added


mandos/src/account.rs
```diff
@@ -61,12 +72,43 @@ impl CheckStorage {
 	}
 }
 
+#[derive(Debug)]
+pub enum CheckEsdt {
```
andrei-marinica, line 76: You can avoid having to use an extra Option, by having DefaultStar, same as with CheckValue. DefaultStar is identical to Star in behaviour, but it occurs when nothing is written in the json. They are separate just so we can reconstruct the mandos the other way around, no other function. DefaultStar would be the Option::None.

elrond-wasm-debug/src/execute_mandos.rs
```diff
@@ -373,6 +437,105 @@ fn check_state(accounts: &mandos::CheckAccounts, state: &mut BlockchainMock) {
 					);
 				}
 			}
+
+			match &expected_account.esdt {
+				Some(CheckEsdt::Equal(eq)) => {
```
andrei-marinica, line 442: There is a Checkable trait, that is used precisely for this. Implementing Checkable for CheckEsdt gives you the "check" method, where you can move the checking logic, with the match and all.
The point is to have less code here, where there is already plenty, and have at least the checking done somewhere else.

mandos/src/account.rs
```diff
@@ -61,12 +72,43 @@ impl CheckStorage {
...
+			),
+		}
+	}
+}
+
+impl CheckEsdt {
+	pub fn is_star(&self) -> bool {
+		matches!(self, CheckEsdt::Star)
+	}
+}
```
andrei-marinica, line 103: You should also implement the Checkable trait here, it takes away code from other, more crowded places.

andrei-marinica (changes requested):

elrond-wasm-debug/src/blockchain_mock.rs
```diff
@@ -181,6 +199,68 @@ impl BlockchainMock {
...
+		let esdt_balance = esdt.get_mut(esdt_token_name).unwrap_or_else(|| {
+			panic!(
+				"Account {} has no esdt tokens with name {}",
+				address_hex(&address),
+				String::from_utf8(esdt_token_name.to_vec()).unwrap()
+			)
+		});
+
+		assert!(
+			*esdt_balance >= *value,
```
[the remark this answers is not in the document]
dorin-iancu: Nope, that gives the following error: 
mismatched types
expected mutable reference `&mut num_bigint::BigUint`
               found reference `&num_bigint::BigUint`

andrei-marinica (approved):

PR #55, opened 2020-12-03
Derive encode decode

dorin-iancu:
Works:
- struct with named fields
- struct with named fields and generics
- struct with unnamed fields (tuple-style structs)
- simple enums
- enums with one or zero fields per value

Doesn't work:
- enums with more than one field per value
- unions


elrond-codec-derive/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+            }
+        },
+        syn::Data::Enum(_) => {
+            let idents = extract_field_names(&ast.data);
+            let value = 0..idents.len() as u8;
+            let name_repeated = std::iter::repeat(name);
+
+            quote! {
+                impl #name {
+                    pub fn to_u8(&self) -> u8 {
```
andrei-marinica, line 263: I would avoid adding extra methods, to avoid clashes with user-defined methods.
The generated code can get duplicated, it's ok because we will never use top_encode in a contract.
I will play some more with traits, for now it can remain like that.

elrond-codec-derive/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+fn impl_top_decode_macro(ast: &syn::DeriveInput) -> TokenStream {
+    let name = &ast.ident;
+    let gen = match &ast.data {
+        syn::Data::Struct(_) => {
+            let (impl_generics, ty_generics, where_clause) = &ast.generics.split_for_impl();
+
+            quote! {
+                impl #impl_generics TopDecode for #name #ty_generics #where_clause {
+                    fn top_decode<I: TopDecodeInput>(input: I) -> Result<Self, DecodeError> {
+                        top_decode_from_nested(input)
```
andrei-marinica, line 301: Long-term, we probably shouldn't assume NestedDecode is implemented, but for now it will be implemented for all structs.

elrond-codec-derive/tests/struct_derive_test.rs
```diff
@@ -0,0 +1,37 @@
...
+#[derive(NestedEncode, NestedDecode, TopEncode, TopDecode)]
+struct StructWithGeneric<ST: SimpleTrait> 
+    where ST:NestedEncode+NestedDecode+TopEncode+TopDecode {
+
+    data: u64,
+    trait_stuff: ST,
+}
+
+#[derive(NestedEncode, NestedDecode, TopEncode, TopDecode)]
+struct TupleStruct(u8, u16, u32);
```
andrei-marinica, line 37: We should probably add some encoding/decoding tests. There are some examples in elrond-codec. You can actually copy SerEx1 entirely (I think it already has some serialization tests).

andrei-marinica (dismissed):

elrond-codec-derive/src/lib.rs
dorin-iancu: Fixed.

andrei-marinica (approved):

PR #57, opened 2020-12-09
More, improved interaction examples

andreibancioiu:


andrei-marinica (approved):

contracts/examples/lottery-egld/interaction/playground.ts
```diff
@@ -0,0 +1,93 @@
+import path = require("path");
```
dorin-iancu, line 1: Is the file extension supposed to be .ts? This looks like Rust code (.rs).

dorin-iancu (approved):

contracts/examples/lottery-egld/interaction/playground.ts
andreibancioiu: Nice that is looks like Rust :)

TypeScript :heart:

dorin-iancu: My bad. Carry on. :+1:

PR #71, opened 2020-12-17
Non payable hook

andrei-marinica:
The branch started as the Arwen nonPayableFuncCheck hook integration, but evolved into a full refactor of the call value processing in SCs.

Contains the following features:
- `#[payable]` annotation more versatile: `#[payable("EGLD")]`,`#[payable("TOKEN-ID")]` `#[payable("*")]`
- `#[payable]` still accepted but throws a warning, will become unsupported in the future.
- `#[payment]` argument attribute now also provides ESDT payment where applicable
- a new TokenIdentifier type that encodes the EGLD special token and any ESDT token
- a new `#[token_identifier]` argument attribute provides the token id. Similar to `#[payment]` it is a fake argument, not exported.
- ABI updated ("payableInTokens" is no longer restricted to "EGLD")
- all new features covered by mandos tests
- async proxies still only accept `#[payable("EGLD")]`, but that is for future updates

 Less visible changes:
- all call value hooks now grouped in a new CallValueApi
- for low-level access, developers now need to write `self.call_value().egld_value()`, etc.
- some optimizations in the handling of call value hooks

Refactoring:
- parse_attr mod was split into a proper folder with many files, since it had grown too large
- an extensive refactoring of elrond-wasm-derive not yet performed, will come soon


elrond-wasm-derive/src/parse_attr/payable_attr.rs
```diff
@@ -0,0 +1,67 @@
+use super::attr_names::*;
+use super::util::*;
+
+static EGLD_DEFAULT: &str = "EGLD";
```
dorin-iancu, line 4: Use &[u8] instead of &str?

elrond-wasm-node/src/api/call_value_api_node.rs
```diff
@@ -0,0 +1,57 @@
...
+		unsafe {
+			let mut name_buffer = [0u8; MAX_POSSIBLE_TOKEN_NAME_LENGTH];
+			let name_len = getESDTTokenName(name_buffer.as_mut_ptr());
+			if name_len == 0 {
+				TokenIdentifier::egld()
+			} else {
+				BoxedBytes::from(&name_buffer[..name_len as usize]).into()
+			}
+		}
+	}
```
dorin-iancu, line 56: Maybe check if the endpoint is payable first? And return an empty token identifier if that's the case.

elrond-wasm-node/src/api/call_value_api_node.rs
```diff
@@ -0,0 +1,57 @@
+use super::ArwenBigUint;
+use crate::ArwenApiImpl;
+use elrond_wasm::api::CallValueApi;
+use elrond_wasm::types::{BoxedBytes, TokenIdentifier};
+
+const MAX_POSSIBLE_TOKEN_NAME_LENGTH: usize = 32;
```
dorin-iancu, line 6: Rename "NAME" to "IDENTIFIER"

elrond-wasm/src/api/call_value_api.rs
```diff
@@ -0,0 +1,58 @@
...
+	/// The method might seem redundant, but there is such a hook in Arwen
+	/// that might be used in this scenario in the future.
+	fn payment_token_pair(&self) -> (BigUint, TokenIdentifier) {
+		let token = self.token();
+		if token.is_egld() {
+			(self.egld_value(), token)
+		} else {
+			(self.esdt_value(), token)
+		}
+	}
```
dorin-iancu, line 57: This is mostly nitpicking, but I usually have this as (identifier, amount) instead of (amount, identifier). Since usually you get the token name, and then the amount when doing checks. Though I can also see the argument of "5 eGLD" instead of "eGLD 5". 

Your call.

elrond-wasm/src/types/token_identifier.rs
```diff
@@ -0,0 +1,194 @@
...
+
+	#[inline]
+	pub fn is_esdt(&self) -> bool {
+		!self.is_egld()
+	}
+
+	#[inline]
+	pub fn into_boxed_bytes(self) -> BoxedBytes {
+		self.0
+	}
```
dorin-iancu, line 36: Rename to "as_boxed_bytes" to keep consistency with the other convert methods.

dorin-iancu (approved):

elrond-wasm-derive/src/contract_gen_callback.rs
```diff
@@ -52,6 +52,9 @@ fn generate_callback_body_regular(methods: &[Method]) -> proc_macro2::TokenStrea
 									ArgMetadata::Payment => {
 										panic!("payment args not allowed in callbacks")
 									},
+									ArgMetadata::PaymentToken => {
+										panic!("payment token args not allowed in callbacks")
+									},
```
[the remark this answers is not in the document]
andrei-marinica: You are right, although this was not always the case.
I would argue that they should be payable by default, which they are anyway.
I will also add the #[payment] and #[payment_token] to callbacks.
They are in dire need of refactoring anyway, it is coming as part of the next big async call push.

elrond-wasm-derive/src/parse_attr/payable_attr.rs
andrei-marinica: This is in the macros not the contract, it is ok to work with string here.

elrond-wasm-node/src/api/call_value_api_node.rs
andrei-marinica: This is done on macro level, for the annotations.
No-one can stop the developer from calling this method, the method also doesn't know whether the endpoint is payable or not.
It will return EGLD if no payment was supplied.

elrond-wasm/src/types/token_identifier.rs
andrei-marinica: `into_boxed_bytes` is actually the correct name here, since it consumes the input (`self` not `&self`)

The convention is outlined here: https://doc.rust-lang.org/1.0.0/style/style/naming/conversions.html

elrond-wasm/src/api/call_value_api.rs
andrei-marinica: Ideally this method should never be called directly by the developer and I've also placed `#[payment_token]` after `#[payment]` everywhere. I also wrote the tests in basic-features like this, with the token afterwards.
I could change everywhere, but I feel it's not worth it right now. Can be done later too.

elrond-wasm-derive/src/parse_attr/payable_attr.rs
andrei-marinica: It might make sense to validate the token names somewhat, so as not to contain funny characters. However, this is I think more relevant when issuing tokens, here we just consume them. Having an invalid token will just cause the smart contract to indirectly not be payable. I'm not fond of having too many restrictions where there is no security vulnerability.

dorin-iancu (approved):

contracts/feature-tests/basic-features/src/lib.rs
```diff
@@ -899,6 +899,144 @@ pub trait BasicFeatures {
...
+			self.call_value().token(),
+			pair_call_value,
+			pair_token_name,
+		)
+			.into()
+	}
+
+	#[endpoint]
+	#[payable("*")]
+	fn payable_any_1(
```
camilbancioiu, line 922: Endpoint names should describe their behavior, in tests as well.

elrond-wasm-debug/src/api/contract_api_mock.rs
```diff
@@ -2,16 +2,21 @@ use super::big_int_api_mock::*;
...
 use crate::async_data::*;
 use crate::{SendBalance, TxContext};
-use alloc::vec::Vec;
 use elrond_wasm::{Address, ArgBuffer, BoxedBytes, CodeMetadata, H256};
 
 impl elrond_wasm::api::ContractHookApi<RustBigInt, RustBigUint> for TxContext {
 	type Storage = Self;
+	type CallValue = Self;
 
 	fn get_storage_raw(&self) -> Self::Storage {
```
camilbancioiu, line 11: Add `#[inline]` as well?

elrond-wasm-derive/src/contract_gen_callback.rs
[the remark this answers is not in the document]
camilbancioiu: Payable callbacks are already supported by the protocol and by Arwen. It must be possible to annotate callbacks as payable (EGLD and ESDT).

elrond-wasm-derive/src/parse_attr/endpoint_attr.rs
```diff
@@ -0,0 +1,126 @@
...
+
+impl ViewAttribute {
+	pub fn parse(m: &syn::TraitItemMethod) -> Option<ViewAttribute> {
+		match find_attr_with_one_opt_token_tree_arg(m, ATTR_VIEW) {
+			None => None,
+			Some(Some(proc_macro2::TokenTree::Ident(ident))) => Some(ViewAttribute {
+				view_name: Some(ident),
+			}),
+			Some(None) => Some(ViewAttribute { view_name: None }),
+			_ => panic!("unexpected endpoint argument tokens"),
```
camilbancioiu, line 57: Having to write `Some(Some(something))` suggests that `find_attr_with_one_opt_token_tree_arg()` should be returning some sort of `Enum`. E.g. an `Enum` containing the values `NoAttribute`, `NullaryAttribute { name-or-something }` instead of `None` and `Some(None)` respectively.

PR #76, opened 2020-12-18
Multisig - clean up storage before performing async_call

andrei-marinica:
See GitHub issue.

### DIFF
--- a/contracts/examples/multisig/src/lib.rs
+++ b/contracts/examples/multisig/src/lib.rs
@@ -343,6 +343,14 @@ pub trait Multisig {
 
 	fn perform_action(&self, action_id: usize) -> SCResult<MultiResultVec<BoxedBytes>> {
 		let action = self.get_action_data(action_id);
+
+		// clean up storage
+		// happens before actual execution, because the async_call kills contract execution,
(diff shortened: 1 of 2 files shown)


contracts/examples/multisig/src/lib.rs
```diff
@@ -350,6 +350,14 @@ pub trait Multisig {
 
 	fn perform_action(&self, action_id: usize) -> SCResult<MultiResultVec<BoxedBytes>> {
 		let action = self.get_action_data(action_id);
+
+		// clean up storage
+		// happens before actual execution, because the async_call kills contract execution,
+		// so cleanup cannot happen afterwards
+		self.set_action_data(action_id, &Action::Nothing);
+		self.set_action_signer_ids(action_id, &[][..]);
```
dorin-iancu, line 358: Use #[storage_clear] instead of manually assigning empty slice

dorin-iancu (approved):

contracts/examples/multisig/src/lib.rs
andrei-marinica: It's ok here, I think the serialization in this case in a no-op.

PR #77, opened 2020-12-18
Cryptokitties

dorin-iancu:
Cryptokitties implementation, split into 3 smart contracts:
1. KittyOwnership - The main contract. It holds all the information regarding who owns which kitty, handles breeding logic, and it performs many other checks. It also implements the ERC721 interface.
2. KittyGeneticAlg - A genetic algorithm mock. This can be easily extended in the future, but I made it as simple as possible for the sake of the example.
3. KittyAuction - Handles the kitty auctioning, where players may either sell their kitties or allow other players to have one of their kitties mate with theirs.


contracts/examples/crypto-kitties/kitty-auction/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+			AsyncCallResult::Err(_) => {
+				// this can only fail if the kitty_ownership contract address is invalid
+				// nothing to revert in case of error
+			},
+		}
+	}
+
+	// private
+
+	fn _create_auction(
```
andrei-marinica, line 275: no need for leading underscores anymore

contracts/examples/crypto-kitties/kitty-genetic-alg/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+#[elrond_wasm_derive::contract(KittyGeneticAlgImpl)]
+pub trait KittyGeneticAlg {
+	#[init]
+	fn init(&self) {}
+
+	// endpoints
+
+	#[endpoint(generateKittyGenes)]
+	fn generate_kitty_genes(&self, matron: Kitty, sire: Kitty) -> SCResult<KittyGenes> {
+		let mut random = Random::new(*self.get_block_random_seed());
```
andrei-marinica, line 17: Isn't the block random seed already random enough?

andrei-marinica, line 17: Isn't the block random seed random enough?
Oh, I see, we need more than 1 number.
If we just need 2 random numbers 1-99, I think we can split the block random seed. (Thinking now, if we shouldn't provide something from the framework that does that).

contracts/examples/crypto-kitties/kitty-genetic-alg/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+		let fur_color_percentage = 1 + random.next_u8() % 99; // val in [1, 100)
+		let matron_fur_color = matron.get_fur_color();
+		let sire_fur_color = sire.get_fur_color();
+		let kitty_fur_color = matron_fur_color.mix_with(
+			&sire_fur_color,
+			fur_color_percentage,
+			100 - fur_color_percentage,
+		);
+
+		let eye_color_percentage = 1 + random.next_u8() % 99; // val in [1, 100)
```
andrei-marinica, line 28: Shouldn't we also somehow include the tx nonce into the randomness? Calling generateKittyGenes twice in the same block could produce 2 identical cats. Is that ok?

andrei-marinica (approved):

contracts/examples/crypto-kitties/kitty-auction/src/lib.rs
dorin-iancu: I prefer having underscores for private functions, but I'm willing to change that if it's not a good convention.

contracts/examples/crypto-kitties/kitty-genetic-alg/src/lib.rs
dorin-iancu: In the current implementation, yes, I could simply take the block random seed and split it, but I prefer having it this way for easier extension later.

sasurobert (changes requested):

sasurobert (approved):

PR #80, opened 2020-12-19
Add ping-pong smart contract

claudiu725:


contracts/examples/ping-pong-egld/mandos/ping-pong-call-ping-second-user.scen.json
```diff
@@ -0,0 +1,76 @@
...
+                },
+                "address:participant2": {
+                    "nonce": "1",
+                    "balance": "2,000,000,000,000",
+                    "storage": {}
+                },
+                "address:the_ping_pong_contract": {
+                    "nonce": "0",
+                    "balance": "1,000,000,000,000",
+                    "storage": {
```
andrei-marinica, line 59: checkState steps are great, but they can be hard to keep up to date when refactoring the storage.
That is why we usually prefer to check the correctness via view functions/sc call outputs.

It's good that you did these kinds of tests, but you might want to avoid being too explicit with the storage for future ones...

... especially since I just requested some storage renames and a small redesign.

contracts/examples/ping-pong-egld/tests/mandos_test.rs
```diff
@@ -0,0 +1,100 @@
+
```
andrei-marinica, line 1: Missing test:
Calling pong_all after at least 1 user already pong-ed.

contracts/examples/ping-pong-egld/mandos/ping-pong-call-ping-before-beginning.scen copy.json
```diff
@@ -0,0 +1,65 @@
...
+    "name": "try to ping before the contract has begun",
+    "steps": [
+        {
+            "step": "externalSteps",
+            "path": "ping-pong-init.scen.json"
+        },
+        {
+            "step": "setState",
+            "accounts": {
+                "address:participant1": {
```
andrei-marinica, line 11: I usually initialize all participants in a steps.json and then just import that in all tests before everything.

andrei-marinica (changes requested):

contracts/examples/ping-pong-egld/tests/mandos_test.rs
claudiu725: Added it.

contracts/examples/ping-pong-egld/mandos/ping-pong-call-ping-before-beginning.scen copy.json
claudiu725: Agreed. Done.

contracts/examples/ping-pong-egld/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+	fn pong_all(&self) -> SCResult<()> {
+		require!(
+			self.get_block_nonce() >= self.get_deadline(),
+			"can't withdraw before deadline"
+		);
+
+		let size = self.get_size();
+		for i in 1..=size {
+			let address = self.get_address(i);
+			self.pong_address(&address).ok();
```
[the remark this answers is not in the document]
claudiu725: Regarding pong_all - it doesn't crash since when any user encounters an error since the SCResult returned by pong (pong_address / pong_by_user_id) in the loop is ignored (note that .ok() is called on the returned value so that the compiler doesn't complain the result is ignored).
I agree with the rest. I changed the mapping to use the user_id instead of the user address (using the naming as well).
The get_or_create_user function proved useful since it allowed me to simplify the code inside the ping method.

contracts/examples/ping-pong-egld/mandos/ping-pong-call-ping-second-user.scen.json
claudiu725: Here I followed the pattern inside the mandos tests for the crowdfunding - and I see the rest of the smart contracts also test the final values by making use of checkState.
checkState is very specific and doesn't allow me to skip fields, hence it's a bit safer than the alternative (testing individual fields). However, I see how it could become a problem for more complex contracts in the future.
For this contract I think checkState is fine since I also implemented a map between (id) <-> (account, status) which I also want to fully test.
If the mapping functionality would have been a module and would have its own tests, if I'm re-using it here then of course, it won't make sense to test its internals.

contracts/examples/ping-pong-egld/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+	#[endpoint]
+	fn pong_all(&self) -> SCResult<()> {
+		require!(
+			self.get_block_nonce() >= self.get_deadline(),
+			"can't withdraw before deadline"
+		);
+
+		let num_users = self.get_num_users();
+		for user_id in 1..=num_users {
+			self.pong_by_user_id(user_id).ok();
```
andrei-marinica, line 104: let _ = self.pong_by_user_id(user_id);
to ignore the result elegantly.
.ok() doesn't do anything here, it's just a getter

contracts/examples/ping-pong-egld/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+	fn pong_by_user_id(&self, user_id: usize) -> SCResult<()> {
+		let user_status = self.get_user_status(user_id);
+		match user_status {
+			UserStatus::New => {
+				sc_error!("can't pong, never pinged")
+			},
+			UserStatus::Registered => {
+				self.set_user_status(user_id, &UserStatus::Withdrawn);
+				let user_address = self.get_user_address(user_id);
+				self.send_tx(&user_address, &self.get_fixed_sum(), "pong");
```
andrei-marinica, line 74: You will need to merge from master, stuff happened since a week ago.
send_tx now asks for &[u8] instead of &str as message, so 
self.send_tx(&user_address, &self.get_fixed_sum(), b"pong");

contracts/examples/ping-pong-egld/wasm/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+lto = true
+debug = false
+panic = "abort"
+
+[dependencies.ping-pong]
+features = [ "wasm-output-mode",]
+path = ".."
+
+[dependencies.elrond-wasm-output]
+version = "0.9.8"
```
andrei-marinica, line 24: it's 0.10.2 now

andrei-marinica: mass replace will be your friend here

contracts/examples/ping-pong-egld/wasm/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "ping-pong-wasm"
+version = "0.0.1"
+authors = [ "you",]
+edition = "2018"
```
andrei-marinica, line 5: we now started adding
publish = false
to all sample contracts

contracts/examples/ping-pong-egld/wasm/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "ping-pong-wasm"
+version = "0.0.1"
+authors = [ "you",]
```
andrei-marinica, line 4: you can add your name here too, or remove the field

contracts/examples/ping-pong-egld/snippets.sh
```diff
@@ -0,0 +1,56 @@
+ALICE="${USERS}/alice.pem"
```
andrei-marinica, line 1: did you get to play around with the snippets?
if yes -> great! we keep it
if no, just copied it from somewhere else -> better leave this out, this contract will be used by other devs as template, better not confuse them, we'll have time for snippets later

contracts/examples/ping-pong-egld/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
```
andrei-marinica, line 1: Last week we added ABI functionality.
ABIs are basically some json files describing the public contract functions.
Anyway, to enable for this contract too just copy the "abi" sub-crate from another contract and change the names (after merge from master). It's similar to how we have a "wasm" sub-crat to every contract.

Also make sure you add your contract to build-wasm.sh.

contracts/examples/ping-pong-egld/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+edition = "2018"
+
+[lib]
+path = "src/lib.rs"
+
+[features]
+wasm-output-mode = ["elrond-wasm-node"]
+
+[dependencies.elrond-wasm]
+version = "0.9.8"
```
andrei-marinica, line 14: mass replace 0.9.8 -> 0.10.2

andrei-marinica (changes requested):

contracts/examples/ping-pong-egld/Cargo.toml
claudiu725: updated to the latest (0.10.3)

contracts/examples/ping-pong-egld/Cargo.toml
claudiu725: done

contracts/examples/ping-pong-egld/snippets.sh
claudiu725: I didn't really use it. I removed it now

contracts/examples/ping-pong-egld/wasm/Cargo.toml
claudiu725: done

contracts/examples/ping-pong-egld/wasm/Cargo.toml
claudiu725: done

contracts/examples/ping-pong-egld/wasm/Cargo.toml
claudiu725: updated to the latest (0.10.3)

contracts/examples/ping-pong-egld/src/lib.rs
claudiu725: I did a merge, on top of this there were a few other changes I needed to add as well in order for the code to compile: deriving from TypeAbi, adding derive_imports!();

contracts/examples/ping-pong-egld/src/lib.rs
claudiu725: done

andrei-marinica (approved):

PR #83, opened 2020-12-21
Cargo clippy linter fixes

andrei-marinica:
Nothing much to say, fixed some linter issues, not all.

It is not definitive, will need to be continued in the future.


elrond-wasm-debug/src/blockchain_mock.rs
```diff
@@ -144,7 +158,7 @@ impl BlockchainMock {
 	fn get_new_address(&self, creator_address: Address, creator_nonce: u64) -> Option<Address> {
 		self.new_addresses
 			.get(&(creator_address, creator_nonce))
-			.map(|addr_ref| addr_ref.clone())
+			.cloned()
```
dorin-iancu, line 161: Typo or intentional? Shouldn't it be .clone() instead of .cloned()?

dorin-iancu (dismissed):

elrond-wasm-debug/src/blockchain_mock.rs
andrei-marinica: .cloned() is the iterator version

andrei-marinica: it means "clone each item from the iterator"

andrei-marinica: clippy proposed this

dorin-iancu (approved):

PR #100, opened 2021-01-05
verifyBls, verifyEd25519, verifySecp256k1 added to Rust framework

dorin-iancu:


contracts/feature-tests/basic-features/src/lib.rs
```diff
@@ -878,6 +878,23 @@ pub trait BasicFeatures {
 		self.keccak256(&input)
 	}
 
+	// Not called, they currently just panic with "Not implemented yet!"
+
+	#[endpoint(verifyBlsSignature)]
+	fn verify_bls_signature(&self, key: Vec<u8>, message: Vec<u8>, signature: Vec<u8>) -> bool {
```
andrei-marinica, line 884: make all arguments &[u8], no need for Vec

elrond-wasm-debug/src/ext_mock.rs
```diff
@@ -444,6 +444,20 @@ impl elrond_wasm::ContractHookApi<RustBigInt, RustBigUint> for TxContext {
 		let hash: [u8; 32] = hasher.result().into();
 		hash.into()
 	}
+
+	// TODO: Remove underscores when implementing
+
+	fn verify_bls(&self, _key: &[u8], _message: &[u8], _signature: &[u8]) -> bool {
+		panic!("Not implemented yet!")
```
andrei-marinica, line 451: also add the method name in the panic message, it makes for easier debugging

elrond-wasm-node/src/ext.rs
```diff
@@ -76,7 +76,10 @@ extern {
     
     // crypto API
     fn sha256(dataOffset: *const u8, length: i32, resultOffset: *mut u8) -> i32;
-    fn keccak256(dataOffset: *const u8, length: i32, resultOffset: *mut u8) -> i32;
+	fn keccak256(dataOffset: *const u8, length: i32, resultOffset: *mut u8) -> i32;
```
andrei-marinica, line 79: cd into elrond-wasm-node and also call `cargo fmt` there to get everything nicely aligned

contracts/feature-tests/basic-features/src/lib.rs
```diff
@@ -878,6 +878,23 @@ pub trait BasicFeatures {
 		self.keccak256(&input)
 	}
 
+	// Not called, they currently just panic with "Not implemented yet!"
```
andrei-marinica, line 881: If you want to add tests, create the mandos files, but do not add `#[test]`-s in Rust. This way we can run them in mandos-go and mandos-rs just ignores them. I did the same for multisig with deployAdder_then_call.scen.json, because contract deployment from contract is not yet implemented in mandos-rs.

contracts/feature-tests/basic-features/src/lib.rs
```diff
@@ -878,6 +878,23 @@ pub trait BasicFeatures {
...
+
+	#[endpoint(verifyBlsSignature)]
```
andrei-marinica, line 883: you can leave the Rust method name to be the endpoint name here, we only use this contract for tests, it's easier to follow this way. I think the other endpoints are snake case too

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #104, opened 2021-01-09
Storage refactor + storage mappers infrastructure

andrei-marinica:
Split the 2 big old contract api traits into several smaller ones: ContractHookApi (main one), CryptoApi, ErrorApi, StorageReadApi, StorageWriteApi.
Moved BigIntApi, BigUintApi with the other APIs.
Also reorganized elrond-wasm-node and elrond-wasm-debug accordingly.
Removed the direct storage methods from ContractHookApi. All storage interaction now goes through get_storage_raw().
Had to finally clean up async-alice and async-bob, they were displaying some very old code. Their mandos tests were also in need for attention.
`#[storage_mapper]` annotation.
Some basic storage mappers: SingleValueMapper, VecMapper, UserMapper. More to come in the future.
Usage of UserMapper in ping-pong-egld, so the feature also has some tests going.
Cleaned up imports in elrond-wasm-debug.
Trait SignalError became identical to ErrorApi, so it was removed.


elrond-wasm/src/storage/mappers/user_mapper.rs
```diff
@@ -0,0 +1,176 @@
...
+	/// Yields the user address for a given id, if the id is valid.
+	pub fn get_user_address(&self, id: usize) -> Option<Address> {
+		let key = self.get_user_address_key(id);
+		// TODO: optimize, storage_load_len is currently called twice
+		if self.api.storage_load_len(key.as_slice()) > 0 {
+			Some(storage_get(self.api.clone(), key.as_slice()))
+		} else {
+			None
+		}
+	}
```
dorin-iancu, line 85: Maybe add a check for id = 0, but it isn't really necessary.

elrond-wasm/src/storage/mappers/vec_mapper.rs
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/// Get item at index from storage.
+	/// Index must be valid (1 <= index <= count).
+	pub fn get(&self, index: usize) -> T {
+		if index == 0 || index > self.count {
+			self.api.signal_error(&b"index out of range"[..]);
+		}
+		storage_get(self.api.clone(), self.get_key(index).as_slice())
+	}
```
dorin-iancu, line 96: I wonder: would it be better to have input indexes be 0-based, but use indexes starting from 1 in the "background"? Since arrays are 0-indexed in Rust, it could create some confusion.

dorin-iancu (dismissed):

elrond-wasm/src/storage/mappers/vec_mapper.rs
andrei-marinica: I think I am actually going to use NonZeroUsize for them. I'm not sure if we should actually shift them, might be confusing to whoever checks the storage directly.
I did think about this too though.

elrond-wasm/src/storage/mappers/user_mapper.rs
andrei-marinica: Should return None for 0, but I didn't get to add tests for that yet. Work in progress.

dorin-iancu (approved):

PR #110, opened 2021-01-27
Add LinkListMapper - a doubly-linked list

claudiu725:


elrond-wasm/src/storage/mappers/linked_list_mapper.rs
```diff
@@ -0,0 +1,240 @@
...
+	fn get_node(&self, index : u32) -> Node {
+		storage_get(self.api.clone(), self.build_index_named_key(NODE_IDENTIFIER, index).as_slice())
+	}
+
+	fn set_node(&self, index : u32, item : Node) {
+		storage_set(self.api.clone(), self.build_index_named_key(NODE_IDENTIFIER, index).as_slice(), &item);
+	}
+
+	fn clear_node(&self, index : u32) {
+		storage_set(self.api.clone(), self.build_index_named_key(NODE_IDENTIFIER, index).as_slice(), &BoxedBytes::empty());
```
andrei-marinica, line 114: storage_set(self.api.clone(), self.build_index_named_key(NODE_IDENTIFIER, index).as_slice(), &[u8; 0][...]);

andrei-marinica: i think

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
```diff
@@ -0,0 +1,240 @@
...
+			self.set_node(node.next, next);
+		}
+
+		self.clear_node(index);
+		self.clear_value(index);
+		self.set_len(len - 1);
+	}
+
+	#[inline]
+    pub fn iter(&self) -> Iter<SA, T> {
```
andrei-marinica, line 196: please call `cargo fmt` to nicely format the code

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
andrei-marinica: I will actually some better base api for mappers, so you can also leave like this for now

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
```diff
@@ -0,0 +1,240 @@
...
+use elrond_codec::{TopDecode, TopEncode};
+use elrond_codec::elrond_codec_derive::{TopDecode, TopEncode};
+
+const NULL_ENTRY : u32 = 0;
+const LEN_IDENTIFIER : &[u8] = b".len";
+const NEW_KEY_IDENTIFIER : &[u8] = b".new";
+const FRONT_IDENTIFIER : &[u8] = b".front";
+const BACK_IDENTIFIER : &[u8] = b".back";
+const NODE_IDENTIFIER : &[u8] = b"*";
+const VALUE_IDENTIFIER : &[u8] = b"&";
```
andrei-marinica, line 15: Please also add a simple mandos test to make the keys obvious.
You can add a method and the test in basic-features.

andrei-marinica (changes requested):

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
claudiu725: Left as-is

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
claudiu725: I enabled the option from VS code to always format on file save. Done.

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
claudiu725: I added it.

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
```diff
@@ -0,0 +1,353 @@
...
+
+	pub fn pop_back(&mut self) -> Option<T> {
+		self.remove_at_index(self.get_back())
+	}
+
+	pub fn pop_front(&mut self) -> Option<T> {
+		self.remove_at_index(self.get_front())
+	}
+
+	pub(crate) fn remove_at_index(&self, index: u32) -> Option<T> {
```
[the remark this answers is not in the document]
andrei-marinica: Might also be worth mentioning in the comment that this is O(1). The rust linked list has the default O(n) remove.

andrei-marinica (changes requested):

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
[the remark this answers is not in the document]
claudiu725: I renamed all usages of `index` to `node_id`, I agree it's more suggestive.
Added rust doc for public functions, with the special O(1) mention for the remove case.
Regarding the `remove_by_node_id` - I would leave it as `pub(crate)` because it would require the user to manage node ids. I would prefer to restrict the API in order to keep the code idiomatic. Managing IDs is usually a sign that another data structure is needed - which should be implemented in this crate.
Note that I added this method because it's going to be used by the follow-up pull request for the SetMapper, where the set is going to keep track of the `node_id`s of each element inserted into the collection. However, it adds the restriction that no duplicate elements are allowed, so they should stay as separate classes.
Reference, for how the remove is used in SetMapper: https://github.com/ElrondNetwork/elrond-wasm-rs/blob/set-and-map-mappers/elrond-wasm/src/storage/mappers/set_mapper.rs#L106

contracts/feature-tests/basic-features/src/lib.rs
```diff
@@ -446,6 +448,32 @@ pub trait BasicFeatures {
...
+	#[endpoint]
+	fn list_mapper_push_back(&self, item: u32) {
+		let mut list_mapper = self.list_mapper();
+		list_mapper.push_back(item);
+	}
+
+	#[endpoint]
+	fn list_mapper_pop_front(&self) {
+		let mut list_mapper = self.list_mapper();
+		list_mapper.pop_front();
```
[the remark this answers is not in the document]
claudiu725: I changed the `list_mapper_pop_front` to return `Option<u32>` and I adapted the mandos test accordingly.
`cargo clippy` doesn't complain about it at all. I think what you're thinking of is the compiler (not clippy) warning that you get when a `Result` (not `Option`) is ignored
```
warning: unused `std::result::Result` that must be used
  --> main.rs:11:5
   |
11 |     returns_result();
   |     ^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_must_use)]` on by default
   = note: this `Result` may be an `Err` variant, which should be handled
```

elrond-wasm/tests/test_linked_list_mapper.rs
```diff
@@ -0,0 +1,67 @@
...
+	assert_eq!(it.next(), Some(43));
+	assert_eq!(it.next(), Some(44));
+	assert_eq!(it.next(), None);
+}
+
+fn assert_list_eq(list: &LinkedListMapper<TxContext, u64>, expected: Vec<u64>) {
+	assert_eq!(list.len(), expected.len());
+	let vec: Vec<u64> = list.iter().collect();
+	assert_eq!(vec, expected);
+}
```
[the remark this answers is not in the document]
claudiu725: I added the checks

andrei-marinica (approved):

PR #116, opened 2021-02-04
Fix mandos checks when calling invalid functions

claudiu725:
I replaced the behavior when an expected field is empty (eg. `status: ""`) - now instead of handling it as a wildcard, it is handled as a default (empty) value. In the case of integers, it's handled as 0.
Added check for when a function is not found. Now the proper error is returned (status: `1`, message: `invalid function (not found)`)
Changed the assertion message to be more like the one returned by the arwen mandos tests: `result code mismatch` and `result message mismatch`, the first one taking priority over the second one.


dorin-iancu (dismissed):

dorin-iancu (approved):

mandos/src/logs.rs
```diff
@@ -27,7 +27,7 @@ impl InterpretableFrom<CheckLogRaw> for CheckLog {
 pub enum CheckLogs {
 	Star,
 	List(Vec<CheckLog>),
-	DefaultStar,
+	DefaultValue,
```
andrei-marinica, line 30: this needs more investigation and testing before we can merge

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #117, opened 2021-02-04
Send esdt API

andrei-marinica:
- created SendApi, which groups all functionality related to sending tokens and interactions with other contracts
- integrated the new TransferESDT hook from Arwen
- added an unsafe buffer for handling values before transfer
- mandos-rs fixes
- contracts now use the new API + more mandos tests


contracts/examples/crowdfunding-esdt/src/lib.rs
```diff
@@ -82,7 +78,8 @@ pub trait Crowdfunding {
 				let esdt_balance = self.get_esdt_balance();
 
 				self.set_esdt_balance(&BigUint::zero());
-				self.pay_esdt(&esdt_token_name, &esdt_balance, &caller);
+				self.send()
+					.direct(&caller, &esdt_token_name, &esdt_balance, &[]);
```
claudiu725, line 82: I think writing `b""` instead of `&[]` would be more suggestive that we're sending a transaction with an empty message.

elrond-wasm/src/api/send_api.rs
```diff
@@ -0,0 +1,71 @@
+use super::BigUintApi;
+use crate::types::{Address, ArgBuffer, BoxedBytes, CodeMetadata, TokenIdentifier};
+
+const DIRECT_ESDT_GAS: u64 = 0;
```
dorin-iancu, line 4: Shouldn't this be "250,000"? https://docs.elrond.com/developers/esdt-tokens/#transfers

dorin-iancu (approved):

contracts/feature-tests/basic-features/mandos/send_esdt.scen.json
```diff
@@ -0,0 +1,116 @@
...
+            }
+        },
+        {
+            "step": "scCall",
+            "txId": "with-message",
+            "tx": {
+                "from": "address:an_account",
+                "to": "address:features_contract",
+                "value": "0",
+                "function": "send_esdt",
```
claudiu725, line 76: There are two calls for `send_esdt` but none for `send_esdt_twice`.
I think `send_esdt_twice` should be tested as well.

contracts/examples/multisig/src/lib.rs
```diff
@@ -492,8 +492,13 @@ pub trait Multisig {
 				for arg in arguments {
 					arg_buffer.push_raw_arg(arg.as_slice());
 				}
-				let new_address =
-					self.deploy_contract(gas_left, &amount, &code, code_metadata, &arg_buffer);
+				let new_address = self.send().deploy_contract(
+					gas_left,
```
camilbancioiu, line 496: Optionally rename `gas_left` to `gas_limit`.

elrond-wasm-node/src/api/send_api_node.rs
```diff
@@ -0,0 +1,211 @@
+use super::ArwenBigUint;
+use crate::ArwenApiImpl;
+use elrond_wasm::api::SendApi;
+use elrond_wasm::types::{Address, ArgBuffer, BoxedBytes, CodeMetadata};
+
+extern "C" {
+	fn transferValue(
+		dstOffset: *const u8,
```
camilbancioiu, line 8: Optionally add `type ExternalPointer = *const u8;` somewhere accessible to all API components, and use it everywhere, if it makes code clearer.

camilbancioiu (approved):

elrond-wasm-node/src/api/send_api_node.rs
```diff
@@ -0,0 +1,211 @@
...
+		argumentsLengthOffset: *const u8,
+		dataOffset: *const u8,
+	);
+}
+
+impl SendApi<ArwenBigUint> for ArwenApiImpl {
+	fn direct_egld(&self, to: &Address, amount: &ArwenBigUint, data: &[u8]) {
+		unsafe {
+			let amount_bytes32_ptr = amount.unsafe_buffer_load_be_pad_right(32);
+			let _ = transferValue(
```
claudiu725, line 74: The return codes of all the C functions here are ignored. Aren't there situations where they may result in errors and should be reported in the rust code (via a panic or transaction error)? Or are they always successful?

elrond-wasm/src/api/send_api.rs
andrei-marinica: Not really, tested on testnet.
It refers to something else - I think you can call a contract with that amount of locked gas. For simple transfers this is not necessary.

contracts/feature-tests/basic-features/mandos/send_esdt.scen.json
andrei-marinica: Only tested it on public testnet, not yet sure if both mandoses can handle it.

elrond-wasm-node/src/api/send_api_node.rs
andrei-marinica: They will kill the execution anyway.

PR #128, opened 2021-02-15
Erc1155

dorin-iancu:
ERC1155 standard contract. Manages both fungible and non-fungible tokens.


contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+		start: &BigUint,
+		end: &BigUint,
+		owner: &Address,
+	) {
+		let big_uint_one = BigUint::from(1u32);
+		let mut token_id = start.clone();
+
+		while &token_id <= end {
+			self.set_token_owner(&type_id, &token_id, &owner);
+
```
sasurobert, line 301: delete empty line

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+	}
+
+	#[view(isApprovedForAll)]
+	fn is_approval_for_all(&self, owner: Address, operator: Address) -> bool {
+		self.get_is_approved(&operator, &owner)
+	}
+
+	// private
+
+	// mock
```
sasurobert, line 255: why do you need a mock ? you have an API from elrondEI.

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+		if self.get_is_fungible(&type_id) == false {
+			let id_first = self.get_last_valid_token_id_for_type(&type_id);
+			let id_last = id_first.clone() + amount;
+
+			self.set_owner_for_range(&type_id, &id_first, &id_last, &creator);
+
+			self.set_last_valid_token_id_for_type(&type_id, &id_last);
+		}
+
+		// self.transfer_single_event(&caller, &from, &to, &id, &amount);
```
sasurobert, line 197: lets have these calls. the even function inside can have 0 implementation for now.

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+				self.increase_balance(&to, &type_id, &amount);
+
+				self.set_token_owner(&type_id, &token_id, &to);
+			}
+		}
+
+		// self.transfer_batch_event(&caller, &from, &to, ids, amounts);
+
+		if self.is_smart_contract_address(&to) {
+			// TODO: async-call
```
sasurobert, line 134: in case of smart contracts you need to set an async call and have the state for increase / decrease in a pending state - thus if the async call fails you should revert those.

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+			self.decrease_balance(&from, &type_id, &amount);
+			self.increase_balance(&to, &type_id, &amount);
+
+			self.set_token_owner(&type_id, token_id, &to);
+		}
+
+		// self.transfer_single_event(&caller, &from, &to, &id, &amount);
+
+		if self.is_smart_contract_address(&to) {
+			// TODO: async-call
```
sasurobert, line 68: in case of smart contracts you need to set an async call and have the state for increase / decrease in a pending state - thus if the async call fails you should revert those.

sasurobert (changes requested):

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+	#[storage_get("isApproved")]
+	fn get_is_approved(&self, operator: &Address, owner: &Address) -> bool;
+
+	#[storage_set("isApproved")]
+	fn set_is_approved(&self, operator: &Address, owner: &Address, is_approved: bool);
+
+	// Events
+
+	/*
+	#[event("0x0000000000000000000000000000000000000000000000000000000000000001")]
```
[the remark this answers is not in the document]
dorin-iancu: We have very minimal support currently. I've left them as comments until some more work is done for them, as currently we can't even return arrays/slices in events.

contracts/examples/erc1155/src/lib.rs
dorin-iancu: It's currently not exposed as a function in the Rust framework, so I'm waiting until the next patch.

contracts/examples/erc1155/src/lib.rs
dorin-iancu: This isn't a function call, this is the event trigger, which as said above, I'll leave them commented until an update is done for them.

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+	#[view(balanceOf)]
+	fn balance_of(&self, owner: &Address, type_id: &BigUint) -> BigUint {
+		self.get_balance_mapper(&owner)
+			.get(&type_id)
+			.unwrap_or_else(|| BigUint::zero())
+	}
+
+	// returns balance for each (owner, id) pair
+	#[view(balanceOfBatch)]
+	fn balance_of_batch(&self, owners: &[Address], type_ids: &[BigUint]) -> Vec<BigUint> {
```
andrei-marinica, line 238: We could also have a version of this that takes VarArgs and return MultiResultVec. Might be a bit easier to use. But it's optional.

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+			.get(type_id)
+			.unwrap_or_else(|| BigUint::zero());
+
+		balance += amount;
+		balance_mapper.insert(type_id.clone(), balance);
+	}
+
+	fn decrease_balance(&self, owner: &Address, type_id: &BigUint, amount: &BigUint) {
+		let mut balance_mapper = self.get_balance_mapper(owner);
+		let mut balance = balance_mapper.get(type_id).unwrap();
```
andrei-marinica, line 282: unwrap panics, please avoid

I think we could use some `get_or_default` method in the mapper, it seems to be often necessary

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+	#[storage_get("isFungible")]
+	fn get_is_fungible(&self, type_id: &BigUint) -> bool;
+
+	#[storage_set("isFungible")]
+	fn set_is_fungible(&self, type_id: &BigUint, is_fungible: bool);
+
+	// last valid id
+
+	#[storage_get("lastValidTypeId")]
+	fn get_last_valid_type_id(&self) -> BigUint;
```
andrei-marinica, line 353: Optional: looks like the VecMapper pattern

andrei-marinica (changes requested):

contracts/examples/erc1155/src/lib.rs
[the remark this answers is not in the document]
andrei-marinica: Will come very soon, it's ok to have them like this for now

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,387 @@
...
+		require!(to != Address::zero(), "Can't transfer to address zero");
+		require!(
+			type_ids.len() == values.len(),
+			"Id and value lenghts do not match"
+		);
+
+		// storage edits are rolled back in case of SCError, 
+		// so the reverting is handled automatically if one of the transfers fails
+		for i in 0..type_ids.len() {
+			let type_id = &type_ids[i];
```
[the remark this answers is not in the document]
dorin-iancu: Didn't know about this actually. Will do!

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+			"Only the token creator may mint more tokens"
+		);
+
+		self.increase_balance(&creator, &type_id, &amount);
+
+		if self.get_is_fungible(&type_id) == false {
+			let id_first = self.get_last_valid_token_id_for_type(&type_id);
+			let id_last = id_first.clone() + amount;
+
+			self.set_owner_for_range(&type_id, &id_first, &id_last, &creator);
```
claudiu725, line 218: Bug: self.set_owner_for_range sets the owner for the range including both start and end, but here the tokens generated are only from start + 1 to end.
For example in the mandos test, if tokens 1-5 already exist, and 3 tokens are added, only the owner for tokens 6, 7, 8 should be set.
Fix: add 1 to let id_first, and subtract one from id_last
Additionally, change the mandos test so that it checks that the owner of the first 5 tokens is untouched (eg. first transfer them to someone else).

claudiu725 (changes requested):

contracts/examples/erc1155/src/lib.rs
sasurobert: I think unwrap makes the wasm code bigger so it should be avoided. if you have get_or_default you can still check if it is the default value or something was written there.

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,527 @@
...
+		);
+		require!(to != Address::zero(), "Can't transfer to address zero");
+		require!(
+			type_ids.len() == values.len(),
+			"Id and value lenghts do not match"
+		);
+
+		// storage edits are rolled back in case of SCError,
+		// so the reverting is handled automatically if one of the transfers fails
+		for (i, type_id) in type_ids.iter().enumerate() {
```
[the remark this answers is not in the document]
dorin-iancu: Great tip!

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+	}
+
+	fn is_valid_token_id(&self, type_id: &BigUint, token_id: &BigUint) -> bool {
+		self.is_valid_type_id(type_id)
+			&& token_id > &0
+			&& token_id <= &self.get_last_valid_token_id_for_type(type_id)
+	}
+
+	fn increase_balance(&self, owner: &Address, type_id: &BigUint, amount: &BigUint) {
+		let mut balance_mapper = self.get_balance_mapper(owner);
```
claudiu725, line 269: increase_balance and decrease_balance can be expressed more simply through balance_of and set_balance

claudiu725 (changes requested):

contracts/examples/erc1155/src/lib.rs
dorin-iancu: That is slightly less efficient, since it has to construct the storage mapper twice

claudiu725: true, although, in my opinion, we should focus on clarity and code re-use first (unless it's performance-critical code - which in this particular case it's not)

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+		require!(to != Address::zero(), "Can't transfer to address zero");
+		require!(
+			type_ids.len() == values.len(),
+			"Id and value lenghts do not match"
+		);
+
+		// storage edits are rolled back in case of SCError,
+		// so the reverting is handled automatically if one of the transfers fails
+		for (type_id, value) in type_ids.iter().zip(values.iter()) {
+			if self.get_is_fungible(type_id) == true {
```
andrei-marinica, line 105: remove ` == true`

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+		Ok(())
+	}
+
+	/// `value` is amount for fungible, token_id for non-fungible
+	#[endpoint(safeBatchTransferFrom)]
+	fn safe_batch_transfer_from(
+		&self,
+		from: Address,
+		to: Address,
+		type_ids: &[BigUint],
```
andrei-marinica, line 85: Optional: if the 2 lists must always have the same length, maybe it would be better to provide that data as a single list of pairs

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+		require!(
+			self.get_caller() == creator,
+			"Only the token creator may mint more tokens"
+		);
+
+		self.increase_balance(&creator, &type_id, &amount);
+
+		if self.get_is_fungible(&type_id) == false {
+			let last_valid_id = self.get_last_valid_token_id_for_type(&type_id);
+			let id_first = last_valid_id.clone() + BigUint::from(1u32);
```
andrei-marinica, line 186: no need for clone, you can add references

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+	#[view(balanceOf)]
+	fn balance_of(&self, owner: &Address, type_id: &BigUint) -> BigUint {
+		self.get_balance_mapper(&owner)
+			.get(&type_id)
+			.unwrap_or_else(|| BigUint::zero())
+	}
+
+	// returns balance for each (owner, id) pair
+	#[view(balanceOfBatch)]
+	fn balance_of_batch(&self, owners: &[Address], type_ids: &[BigUint]) -> SCResult<Vec<BigUint>> {
```
andrei-marinica, line 232: Optional: can also provide them as argument pairs, i.e.
VarArgs<MultiArg2<Address, BigUint>>
might be easier to work with, but purely optional

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+	fn get_token_type_creator(&self, type_id: &BigUint) -> Address;
+
+	#[storage_set("tokenTypeCreator")]
+	fn set_token_type_creator(&self, type_id: &BigUint, creator: &Address);
+
+	// token type uri
+
+	#[view(getTokenTypeUri)]
+	#[storage_get("tokenTypeUri")]
+	fn get_token_type_uri(&self, type_id: &BigUint) -> Vec<u8>;
```
andrei-marinica, line 482: can be BoxedBytes, since it never grows

andrei-marinica (approved):

contracts/examples/erc1155/src/lib.rs
dorin-iancu: I'll leave them like this for now, it's a bit clearer and more consistent with the transfer single endpoint, but maybe I'll change at some point. Good tip though!

claudiu725 (approved):

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,545 @@
...
+
+elrond_wasm::imports!();
+elrond_wasm::derive_imports!();
+
+const INTERFACE_SIGNATURE_ERC165: u32 = 0x01ffc9a7;
+const INTERFACE_SIGNATURE_ERC1155: u32 = 0xd9b67a26;
+
+const ON_ERC_RECEIVED_ENDPOINT_NAME: &[u8] = b"onERC1155Received";
+const ON_ERC_BATCH_RECEIVED_ENDPOINT_NAME: &[u8] = b"onERC1155BatchReceived";
+const ACCEPTED_TRANSFER_ANSWER: u32 = 0xbc197c81;
```
sasurobert, line 14: why is this needed ?

contracts/examples/erc1155/src/lib.rs
```diff
@@ -0,0 +1,545 @@
+#![no_std]
+
+use elrond_codec::test_util::top_encode_to_vec_or_panic;
+use elrond_wasm::{HexCallDataSerializer, MultiArg2};
+
+elrond_wasm::imports!();
+elrond_wasm::derive_imports!();
+
+const INTERFACE_SIGNATURE_ERC165: u32 = 0x01ffc9a7;
```
sasurobert, line 9: why are these consts needed ?

contracts/examples/erc1155/src/lib.rs
dorin-iancu: removed

andrei-marinica (approved):

sasurobert (dismissed):

sasurobert (approved):

PR #132, opened 2021-02-20
async call redesign - part 2

andrei-marinica:
Added support for transfer-execute hooks.
Syntax update, the highlight being:
```
		contract_call!(self, to, VaultProxy) // -> creates a ContractProxy
			.retrieve_funds(token, payment) // -> creates a ContractCall with serialized call data
			.async_call() // -> converts to an AsyncCall object, replace this line to get transfer-execute
			.with_callback(self.callbacks().retrieve_funds_callback()) // -> also add a callback programatically, this saves the closure data
```
Also some minor imports cleanup.

### DIFF
--- a/contracts/examples/crypto-kitties/kitty-auction/src/lib.rs
+++ b/contracts/examples/crypto-kitties/kitty-auction/src/lib.rs
@@ -71,16 +71,6 @@ pub trait KittyAuction {
 		}
 	}
 
-	#[endpoint]
-	fn claim(&self) -> SCResult<()> {
-		only_owner!(self, "Only owner may call this function!");
-
-		self.send()
-			.direct_egld(&self.get_caller(), &self.get_sc_balance(), b"claim");
-
-		Ok(())
-	}
-
 	// views
 
 	#[view(isUpForAuction)]
--- a/contracts/feature-tests/basic-features/mandos/get_caller.scen.json
+++ b/contracts/feature-tests/basic-features/mandos/get_caller.scen.json
@@ -1,0 +1,58 @@
+{
+    "name": "count ones",
+    "comment": "should fail if the processor doesn't support the `count ones` instruction",
+    "steps": [
+        {
+            "step": "setState",
+            "accounts": {
+                "address:features_contract": {
+                    "nonce": "0",
+                    "balance": "0",
+                    "storage": {},
+                    "code": "file:../output/basic-features.wasm"
+                },
+                "address:an_account": {
+                    "nonce": "0",
+                    "balance": "0",
+                    "storage": {},
+                    "code": ""
+                }
+            }
+        },
+        {
+            "step": "scCall",
+            "txId": "get_caller",
+            "tx": {
+                "from": "address:an_account",
+                "to": "address:features_contract",
+                "value": "0",
+                "function": "get_caller",
+                "arguments": [],
+                "gasLimit": "0x100000",
+                "gasPrice": "0"
+            },
+            "expect": {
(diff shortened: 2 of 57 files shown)


contracts/feature-tests/basic-features/mandos/get_caller.scen.json
```diff
@@ -0,0 +1,58 @@
+{
+    "name": "count ones",
+    "comment": "should fail if the processor doesn't support the `count ones` instruction",
```
claudiu725, line 3: Update name/comment

contracts/feature-tests/basic-features/mandos/get_caller.scen.json
```diff
@@ -0,0 +1,58 @@
...
+                "gas": "*",
+                "refund": "*"
+            }
+        },
+        {
+            "step": "scQuery",
+            "txId": "get_caller",
+            "tx": {
+                "to": "address:features_contract",
+                "function": "get_prev_block_timestamp",
```
claudiu725, line 47: Shouldn't this second query should test get_gas_left_endpoint, not get_prev_block_timestamp?

claudiu725 (changes requested):

elrond-wasm/src/types/interaction/contract_call.rs
```diff
@@ -1,84 +1,115 @@
...
+	pub fn transfer_execute(self) -> TransferExecute<BigUint> {
+		TransferExecute {
+			to: self.to,
+			token: self.token,
+			amount: self.payment,
+			endpoint_name: self.endpoint_name,
+			arg_buffer: self.arg_buffer,
+			gas_limit: 0,
+		}
+	}
```
dorin-iancu, line 114: Wouldn't it be a bit better to have a single TransferExecute struct? Having TransferEgldExecute and TransferEsdtExecute seems a bit overkill. This is entirely optional and you probably have a good reason for splitting them like this, but give it a thought, at least.

dorin-iancu (approved):

contracts/examples/crypto-kitties/kitty-auction/src/lib.rs
```diff
@@ -71,16 +71,6 @@ pub trait KittyAuction {
 		}
 	}
 
-	#[endpoint]
-	fn claim(&self) -> SCResult<()> {
```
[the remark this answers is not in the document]
dorin-iancu: I asked him to do it :) The endpoint isn't needed at all for that contract, and it could lead to owner "stealing" auction locked funds.

Copy-paste oversight on my part when writing those contracts.

claudiu725:
I think some changes from the scQuery pull request may have leaked into this one

contracts/examples/crypto-kitties/kitty-auction/src/lib.rs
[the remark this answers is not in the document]
andrei-marinica: Dorin kindly asked me. The owner is not supposed to be able to do this.

elrond-wasm/src/types/interaction/contract_call.rs
andrei-marinica: If the choice between the hooks is done at runtime, the compiler cannot know which one will be called and both end up in the binary. The only way to separate them at compile time is to have separate structs (or another level of generics, but that's just overkill, I feel). I tried to keep the structs absolutely minimal. The relevant logic is in the conversion methods in ContractCall.

claudiu725 (approved):

PR #158, opened 2021-03-03
elrond-wasm-derive major refactoring

andrei-marinica:
Reorganized modules.
Changed the model: separated method annotations:
- public role - refers to how a method appears to the "outside world"
- auto-implementations - refers to how a method gets implemented


elrond-wasm-derive/src/model/endpoint.rs
```diff
@@ -0,0 +1,29 @@
...
+#[derive(Clone, Debug)]
+pub struct EndpointMetadata {
+	pub public_name: syn::Ident,
+	pub payable: MethodPayableMetadata,
+}
+
+/// Method visibility from the point of view of the smart contract
+#[derive(Clone, Debug)]
+pub enum PublicRole {
+	/// The smart contract constructor. Ther ecan be only one.
```
dorin-iancu, line 17: typo

dorin-iancu (approved):

PR #160, opened 2021-03-04
egld-esdt-swap contract, ESDT issue/mint

andrei-marinica:
- copied egld-esdt-swap contract from own repo to framework
- added ESDT issue/mint calls directly from framework
- egld-esdt-swap minor storage mappers refactoring
- egld-esdt-swap event logs


dorin-iancu (approved):

elrond-wasm/src/esdt.rs
```diff
@@ -0,0 +1,169 @@
...
+		ESDTSystemSmartContractProxy {
+			_phantom: core::marker::PhantomData,
+		}
+	}
+
+	/// Produces a contract call to the ESDT system SC,
+	/// which causes it to issue a new ESDT token.
+	pub fn issue(
+		&self,
+		issue_cost: BigUint,
```
dorin-iancu, line 35: Since the issue cost is always 5 eGLD, maybe have it as a constant in this file?

elrond-wasm/src/esdt.rs
```diff
@@ -0,0 +1,169 @@
...
+
+		contract_call.push_argument_raw_bytes(&b"canUpgrade"[..]);
+		contract_call.push_argument_raw_bytes(bool_name_bytes(can_upgrade));
+
+		contract_call
+	}
+
+	/// Produces a contract call to the ESDT system SC,
+	/// which causes it to issue a new ESDT token.
+	pub fn mint(&self, token_name: &[u8], amount: &BigUint) -> ContractCall<BigUint> {
```
dorin-iancu, line 86: Rename token_name to token_identifier. Same for the rest of the functions.

dorin-iancu (approved):

elrond-wasm/src/esdt.rs
andrei-marinica: The issue cost will almost certainly not stay 5 EGLD. I would actually remove the number from the contract altogether and let the protocol validate. Otherwise, a protocol update might leave the contract broken

elrond-wasm/src/esdt.rs
andrei-marinica: esdt_token_identifier is even better I think
only esdt allowed here

claudiu725 (dismissed):

contracts/examples/egld-esdt-swap/src/lib.rs
```diff
@@ -0,0 +1,230 @@
+#![no_std]
+
+elrond_wasm::imports!();
+elrond_wasm::derive_imports!();
+
+const ESDT_ISSUE_COST: u64 = 5000000000000000000; // 5 EGLD
```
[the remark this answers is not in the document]
andrei-marinica: yeah, actually doesn't belong anywhere in a contract or in the framework, the protocol will validate this

elrond-wasm/src/esdt.rs
claudiu725: If we leave it up to the protocol to do the validation - and the validation fails, are the funds returned to the smart contract or to the user which triggered the call to the issue?

andrei-marinica: good point, we need to check

dorin-iancu (approved):

claudiu725 (approved):

PR #166, opened 2021-03-08
Mandos biguint, nested

claudiu725:
Add mandos "biguint:" type specifier, which can be used when dealing with **nested** BigUint values.
The "biguint:" specifier serializes the number N as an u32, which represents the number of bytes in the binary representation of the value. The u32 is followed by N bytes representing the actual value.
Example usage:
"biguint:1234" (equivalent to "u32:2|1234")
"biguint:0x112233" (equivalent to "u32:3|0x112233")
Add "nested:", which is identical in behavior, but should be used for nested byte arrays (BoxedBytes, Vec<u8>, [u8] etc.)
Example usage:
"nested:0x112233" (equivalent to "u32:3|0x112233")

### DIFF
--- a/mandos/src/value_interpreter.rs
+++ b/mandos/src/value_interpreter.rs
@@ -11,6 +11,7 @@ const FILE_PREFIX: &str = "file:";
 const KECCAK256_PREFIX: &str = "keccak256:";
 
 const BIGUNT_PREFIX: &str = "biguint:";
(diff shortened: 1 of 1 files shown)


andrei-marinica:
Already implemented in https://github.com/ElrondNetwork/elrond-wasm-rs/pull/164

PR #167, opened 2021-03-08
Mandos rs check logs

dorin-iancu:
Checking logs in mandos-rs


elrond-wasm/src/types/interaction/arg_buffer.rs
```diff
@@ -34,6 +34,16 @@ impl ArgBuffer {
 		self.arg_data.as_ptr()
 	}
 
+	/// returns a clone of the raw arg data
+	pub fn arg_data(&self) -> Vec<u8> {
```
andrei-marinica, line 38: Better return as slice &[u8] and let the whoever uses the method to .clone() is necessary

elrond-wasm/src/types/interaction/arg_buffer.rs
```diff
@@ -34,6 +34,16 @@ impl ArgBuffer {
...
+		self.arg_data.clone()
+	}
+
+	/// returns a clone of the raw arg data lengths
+	pub fn arg_lengths(&self) -> Vec<usize> {
```
andrei-marinica, line 43: Better return as slice &[usize] and let the whoever uses the method to .clone() is necessary

elrond-wasm-debug/src/execute_mandos.rs
```diff
@@ -377,6 +381,42 @@ fn check_tx_output(tx_id: &str, tx_expect: &TxExpect, tx_result: &TxResult) {
...
+					expected_log.topics.iter().map(|topic| verbose_hex(&topic.value)).collect::<String>(),
+					verbose_hex(&expected_log.data.value),
+					address_hex(&actual_log.address),
+					vec_u8_to_string(&actual_log.identifier),
+					actual_log.topics.iter().map(|topic| verbose_hex(&topic)).collect::<String>(),
+					verbose_hex(&actual_log.data),
+				);
+			}
+		},
+		CheckLogs::DefaultStar => assert!(
```
andrei-marinica, line 413: In mandos-go unspecified logs are treated as "*", but leave it like this for now, I'll go over all mandos defaults very soon, they are not properly in sync between go and rust.

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #174, opened 2021-03-11
core Result can be used as endpoint result

andrei-marinica:


dorin-iancu (dismissed):

claudiu725 (approved):

elrond-wasm-derive/src/generate/method_call_gen.rs
```diff
@@ -112,7 +112,23 @@ pub fn generate_body_with_result(
...
+			// Because of Rust's orphan rules, Result cannot be made to implement EndpointResult.
+			// To still allow developers to use it as an endpoint result,
+			// we set up a manual conversion via macro magic.
+			// We still let the EndpointResult trait to do the heavy lifting.
+			if let syn::Type::Path(type_path) = ty.as_ref() {
+				let type_path_segment = type_path.path.segments.last().unwrap();
+				let type_str = type_path_segment.ident.to_string();
+				if type_str == "Result" {
+					return quote! {
+						let result = elrond_wasm::types::SCResult::from_result(#mbody);
```
claudiu725, line 125: Are name collisions possible with another local variable named result?

claudiu725 (approved):

elrond-wasm-derive/src/generate/method_call_gen.rs
andrei-marinica: SCResult::from_result only accepts core::Result as argument, it would fail to compile that

PR #175, opened 2021-03-15
Multisig storage mappers & real gas

andrei-marinica:
- refactoring: storage mappers
- real gas tests


elrond-wasm/src/storage/mappers/user_mapper.rs
```diff
@@ -136,6 +143,29 @@ where
 		user_id
 	}
 
+	/// Tries to insert a number of addresses.
+	/// Calls a lambda function for each, with the new user id and whether of nor the user was already present.
```
dorin-iancu, line 147: nor -> not

dorin-iancu (approved):

claudiu725 (approved):

PR #176, opened 2021-03-15
Local esdt and nft api

dorin-iancu:
Implementation of Local Mint and Local Burn for Fungible (i.e. the legacy ESDT token type) hooks.
Implementation of ESDT as NFT API function hooks (issue, create, transfer etc.)


sasurobert (changes requested):
looks good. needs implementation of bigIntGetESDTEXternalBalance and transferCreateRole

sasurobert (dismissed):
looks good. needs implementation of bigIntGetESDTEXternalBalance and transferCreateRole

elrond-wasm/src/esdt.rs
```diff
@@ -29,12 +33,112 @@ impl<BigUint: BigUintApi> ESDTSystemSmartContractProxy<BigUint> {
...
+	}
+
+	/// Produces a contract call to the ESDT system SC,
+	/// which causes it to issue a new non-fungible ESDT token.
+	pub fn issue_non_fungible(
+		&self,
+		issue_cost: BigUint,
+		token_display_name: &BoxedBytes,
+		token_ticker: &BoxedBytes,
+		can_freeze: bool,
```
andrei-marinica, line 78: We should probably group all these arguments in a ESDTMetadata struct, but it is ok for now

andrei-marinica (dismissed):

andrei-marinica (approved):

PR #178, opened 2021-03-16
Managed buffers old implementation [ABANDONED]

andrei-marinica:
work in progress


andrei-marinica:
This is a very old initial attempt at managed buffers on the old API model. Now abandoned.

PR #181, opened 2021-03-16
executeOnDestContext - retrieve raw results

andrei-marinica:
Low-level functionality to extract results after an executeOnDestContext call.

### DIFF
--- a/elrond-wasm-node/src/api/send_api_node.rs
+++ b/elrond-wasm-node/src/api/send_api_node.rs
@@ -79,7 +79,7 @@ extern "C" {
 		argumentsLengthOffset: *const u8,
 		dataOffset: *const u8,
 	) -> i32;
-	
+
 	fn executeOnDestContextByCaller(
 		gas: u64,
 		addressOffset: *const u8,
@@ -90,7 +90,7 @@ extern "C" {
 		argumentsLengthOffset: *const u8,
 		dataOffset: *const u8,
 	) -> i32;
-	
+
 	fn executeOnSameContext(
 		gas: u64,
 		addressOffset: *const u8,
@@ -339,16 +339,8 @@ impl SendApi<ArwenBigUint> for ArwenApiImpl {
 }
 
 /// Retrieves already pushed results, via `finish`.
-unsafe fn get_return_data(return_index: i32) -> BoxedBytes {
-	let len = getReturnDataSize(return_index);
-	let mut res = BoxedBytes::allocate(len as usize);
-	if len > 0 {
-		let _ = getReturnData(return_index, res.as_mut_ptr());
-	}
-	res
-}
-
-/// Retrieves already pushed results, via `finish`.
+/// `from_index` is inclusive.
+/// `to_index` is exclusive.
 unsafe fn get_return_data_range(from_index: i32, to_index: i32) -> Vec<BoxedBytes> {
 	let num_results = to_index - from_index;
 	let mut result = Vec::with_capacity(num_results as usize);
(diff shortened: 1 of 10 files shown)


elrond-wasm/src/types/interaction/arg_buffer.rs
```diff
@@ -67,6 +68,16 @@ impl ArgBuffer {
 	}
 }
 
+impl From<&[BoxedBytes]> for ArgBuffer {
+	fn from(raw_args: &[BoxedBytes]) -> Self {
+		let mut result = ArgBuffer::new();
```
camilbancioiu, line 73: Suggest renaming `result` to `arg_buffer` for clarity.

elrond-wasm/src/types/interaction/arg_buffer.rs
```diff
@@ -67,6 +68,16 @@ impl ArgBuffer {
 	}
 }
 
+impl From<&[BoxedBytes]> for ArgBuffer {
```
camilbancioiu, line 71: :+1:

elrond-wasm-node/src/api/send_api_node.rs
```diff
@@ -310,12 +328,34 @@ impl SendApi<ArwenBigUint> for ArwenApiImpl {
...
 			&ArwenBigUint::from(0u32),
 			function,
 			&arg_buffer,
-		)
+		);
+	}
+}
+
+/// Retrieves already pushed results, via `finish`.
+unsafe fn get_return_data(return_index: i32) -> BoxedBytes {
```
camilbancioiu, line 342: Suggest reordering the function definitions: `get_return_data_range()` comes first, and `get_return_data()` comes second.

elrond-wasm-node/src/api/send_api_node.rs
```diff
@@ -310,12 +328,34 @@ impl SendApi<ArwenBigUint> for ArwenApiImpl {
...
 			&own_address,
 			&ArwenBigUint::from(0u32),
 			function,
 			&arg_buffer,
-		)
+		);
+	}
+}
+
+/// Retrieves already pushed results, via `finish`.
```
camilbancioiu, line 341:
```suggestion
/// Retrieves already pushed individual result at given index, via `finish`.
```

elrond-wasm-node/src/api/send_api_node.rs
```diff
@@ -310,12 +328,34 @@ impl SendApi<ArwenBigUint> for ArwenApiImpl {
...
+	}
+	res
+}
+
+/// Retrieves already pushed results, via `finish`.
+unsafe fn get_return_data_range(from_index: i32, to_index: i32) -> Vec<BoxedBytes> {
+	let num_results = to_index - from_index;
+	let mut result = Vec::with_capacity(num_results as usize);
+	if num_results > 0 {
+		for index in from_index..to_index {
```
camilbancioiu, line 356: Completely optional, but it might be more readable to use `from_index..=to_index` here, and call `get_return_data_range(num_return_data_before, num_return_data_after - 1)`. Variable name `to_index` suggests index-up-to-and-inclusively, but it is not used as such.

In case you make this change, `let num_results = to_index - from_index + 1` will be needed above.

andrei-marinica: the convention "inclusive lower bound, exclusive upper bound" is quite pervasive, it is idiomatic of Rust and other languages, and in fact quite easy to work with. Even the simple `for` below has it. I am, however in favour of a better name

andrei-marinica: or a comment, I think it might be enough here

camilbancioiu (approved):

PR #191, opened 2021-03-29
Add `entry` syntax for maps (and some other improvements)

claudiu725:
- implement `entry` syntax for `MapMapper` and `MapStorageMapper` for managing entries in-place
  - inspired by [std::collections::hash_map::HashMap::entry](https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.entry)
  - however, the implementation the doesn't use the `std` library
  - it differs from the standard implementation above when it comes to mutability:
    - `OccupiedEntry::update` replaces [OccupiedEntry::get_mut](https://doc.rust-lang.org/std/collections/hash_map/struct.OccupiedEntry.html#method.get_mut) since the value has to be written back to storage - this pattern is similar to the one in `SingleValueMapper`
    - for `MapStorageMapper` the API of `Entry` is restricted even further, since the mapped values are storage mappers themselves and have some limitations:
      - only default-constructed instances are allowed
      - the ownership of the nested storage mapper can't be transfered outside (as it would imply copying the entire storage of the value)
- `require!` and `sc_error!` statements are now compatible with `Result<T, SCError>` return values
  - this is done by implicitly converting the result type inside the macro
  - `SCResult` and `Result<T, SCError>` are the only ones supported at the moment
  - the `Result<T, SCError>` type is useful since it allows integration with [`Option::ok_or`](https://doc.rust-lang.org/core/option/enum.Option.html#method.ok_or) and [`Option::ok_or_else`](https://doc.rust-lang.org/core/option/enum.Option.html#method.ok_or_else) such as `value.ok_or("example error")?;`
- `require!` statements can now be used in the functions passed as arguments to `SingleValueMapper::update`
  - this is also implemented for the newly added `OccupiedEntry::update` for both `MapMapper` and `MapStorageMapper`
- add the `Default` trait for `BigUint`, since the use case for managing a balance with a default of 0 is fairly common
  - it simplifies the code somewhat, allowing a smart contract developer to write `example.unwrap_or_default();` instead of `example.unwrap_or(&BigUint::zero());` (or some variant of this) - where `example` is of type `Option<BigUint>`
  - this also applies to maps (MapMapper, MapStorageMapper), where `Entry::or_default` can be used
- improve the error messages of mandos-rs when various inputs aren't valid by also including the input string which generated the error (eg. for `Account not found` / `Unknown contract`) - useful in larger mandos tests where the failure point might not be obvious
- error message fix (`unexpected view argument tokens`)
- fix gas costs for the map_storage_mapper mandos tests


elrond-wasm/src/storage/mappers/map_storage_mapper.rs
```diff
@@ -227,3 +235,155 @@ where
...
+
+	/// Syntactic sugar, to more compactly express a get, update and set in one line.
+	/// Takes whatever lies in storage, apples the given closure and saves the final value back to storage.
+	/// Propagates the return value of the given function.
+	pub fn update<R, F: FnOnce(&mut V) -> R>(&mut self, f: F) -> R {
+		let mut value = self.get();
+		f(&mut value)
+	}
+
+	/// Takes the value of the entry out of the map, removing it.
```
dorin-iancu, line 385: Comment sounds a bit weird. Maybe rephrase to "Removes the entry from the map" or similar

dorin-iancu (approved):

elrond-wasm/src/storage/mappers/map_storage_mapper.rs
claudiu725: Done

andrei-marinica (approved):

PR #202, opened 2021-04-12
OperationCompletionStatus minor improvements

andrei-marinica:


dorin-iancu (approved):
Maybe also implement the encoding/decoding traits through derives? Might be useful in some very specific scenarios, but it's your call.

sasurobert (approved):

PR #206, opened 2021-04-18
Added return value to send api functions

catalinnnn:
It has been made a change in Arwen such that when a smart contract A calls direct esdt execute with function call to smart contract B and that function returns sc_error, the execution will continue in A. So return values from send api functions will be needed to check if the function call returned with success or not. Return value of 0 indicates success and != 0 indicates failure. Tested on DEX smart contracts.

### DIFF
--- a/elrond-wasm/src/api/send_api.rs
+++ b/elrond-wasm/src/api/send_api.rs
@@ -17,7 +17,7 @@ where
 {
 	/// Sends EGLD to a given address, directly.
 	/// Used especially for sending EGLD to regular accounts.
-	fn direct_egld(&self, to: &Address, amount: &BigUint, data: &[u8]) -> i32;
+	fn direct_egld(&self, to: &Address, amount: &BigUint, data: &[u8]);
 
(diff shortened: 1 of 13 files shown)


elrond-wasm/src/api/send_api.rs
```diff
@@ -39,8 +39,8 @@ where
 		token: &[u8],
 		amount: &BigUint,
 		data: &[u8],
-	) {
-		self.direct_esdt_execute(to, token, amount, 0, data, &ArgBuffer::new());
+	) -> i32 {
```
andrei-marinica, line 42: This is a hack, we probably shouldn't copy the pattern to all API methods, but I'll allow it for now.

andrei-marinica (dismissed):

sasurobert (dismissed):

dorin-iancu (approved):

dorin-iancu (approved):

andrei-marinica (approved):

PR #213, opened 2021-04-30
Submodules, call proxy and API trait redesign

andrei-marinica:
### Major redesign of important framework components:

- The arguments to contract/module/proxy annotations are gone. All items are generated in the same Rust module. Both submodule inclusion and contract calls are now Rust-module-aware.
- Submodule imports are now expressed as supertraits instead of the module getter annotated methods. Note: explicitly specifying the Rust module is required, in order for the framework to fetch generated types and functions from that module.
- Each contract now generates its own callable proxy to ease calling it. Caller contracts do no longer need to define a call interface, they can import it from the crate of the contract they want to call. Callable proxies contain the methods from the main contract, as well as from all the modules. Note: calling a contract requires the caller to specify the Rust module where it resides.
- We no longer have a separate syntax/parser/code generation for call proxies. They are just contracts with no implementations and annotated with `#[elrond_wasm_derive::proxy]` instead of `#[elrond_wasm_derive::contract]`.
- BigUint and BigInt are now associated types instead of generics in all API traits. Contracts need to specify them as `Self::BigUint` instead of just `BigUint`. Although more verbose, this might be more intuitive for the developer.
- `ContractCall`s, `AsyncCall`s and all other call & transfer result types now contain a reference to the Send API. This also means the `execute_on_dest_context` method no longer requires an api argument.
- Callbacks and callback proxies needed to be adapted to the new system, but work similar to how they did in the past.
- Callbacks can now have names, just like endpoints. This name gets saved in the callback closure in storage, but has no other impact on the contract. The reason I needed it was to help me with defining callback forwarders and avoiding some name collisions there. Callback forwarders are still needed for a little longer, until module callbacks are properly implemented.

### Still not done:
- Callbacks still don't cascade to modules. This will follow soon.
- A proxy getter method annotation in the caller contract could help clean up the syntax a bit. The syntax is bearable, but not great. It could also help us get rid of having to write `use called_contract_mod::Proxy as _;` before each contract that calls another.
- Some of the code generator code is in need of refactoring (especially the function that puts together all the the contract pieces in the end). There are a lot of pieces and we use a flag to keep some of them off modules, which is not a great pattern.
- It should be easy to add an AsyncCall `execute` method now. This will become the norm when we get Arwen's multi-asyncs.
- To investigate: We currently write `AsyncCall<Self::SendApi>`. We could make it work with `Self::AsyncCall`, but it might not be worth it.

### Notes and some limitations:
- Unlike in the past, only 1 contract/module/proxy is allowed per Rust module. Use `mod { ... }` to separate them if they are in the same file.
- The `callbacks()` method has issues with modules. I managed to somewhat avoid the problem by not generating it at all for modules that have no callbacks. If both a module and its parent have it, though, we get a name conflict, which is fine compile-wise, but when calling it, it needs to be disambiguated with some pretty hairy syntax.
- All modules of a contract need to appear in the main contract trait, even if they are already referenced indirectly in another module. For future investigation: private modules that auto-implement their dependencies. Auto-implementations for sub-sub-modules is now possible but it leads to conflicts if 2 modules implement the same module. We might come up with a special annotation at some point, but it is unclear if it worth it. Low prio in any case.
- The old module getter annotation still exists, but can no longer be used.


dorin-iancu (dismissed):

elrond-wasm/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![no_std]
 #![feature(never_type)]
+// #![feature(generic_associated_types)]
```
sasurobert, line 3: could be deleted

sasurobert (commented):
superb job!

dorin-iancu (approved):

claudiu725 (approved):

dorin-iancu (approved):

PR #215, opened 2021-05-05
added esdt transfer and send functions for any ESDT

catalinnnn:
Identified this as being a case where a lot of contracts might do the handle independently, each on their own.
Think it's best we do the handling in the framework.

### DIFF
--- a/elrond-wasm/src/api/send_api.rs
+++ b/elrond-wasm/src/api/send_api.rs
@@ -77,7 +77,8 @@ where
 	}
 
 	/// Sends ESDT tokens to the target address. Handles any type of ESDT.
-	fn transfer_esdt(&self, token: &TokenIdentifier, nonce: u64, amount: &BigUint, to: &Address) {
+	/// Note: this does not work with EGLD, use only with ESDT.
+	fn transfer_tokens(&self, token: &TokenIdentifier, nonce: u64, amount: &BigUint, to: &Address) {
 		if amount > &0 {
 			if nonce == 0 {
 				let _ =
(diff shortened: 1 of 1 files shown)


elrond-wasm/src/api/send_api.rs
```diff
@@ -76,6 +76,24 @@ where
 		}
 	}
 
+	/// Sends ESDT tokens to the target address. Handles any type of ESDT.
+	fn transfer_esdt(&self, token: &TokenIdentifier, nonce: u64, amount: &BigUint, to: &Address) {
```
claudiulataretu, line 80: I would rather call this method transfer_tokens. You are handling both ESDTs and NFTs in the same method

elrond-wasm/src/api/send_api.rs
```diff
@@ -277,6 +295,17 @@ where
 		self.call_local_esdt_built_in_function(gas, b"ESDTNFTBurn", &arg_buffer);
 	}
 
+	/// Burns ESDT tokens. Handles any type of ESDT.
+	fn burn_esdt(&self, token: &TokenIdentifier, nonce: u64, amount: &BigUint, gas: u64) {
```
claudiulataretu, line 299: I would rather call this method burn_tokens. You are handling both ESDTs and NFTs in the same method

claudiulataretu (changes requested):

elrond-wasm/src/api/send_api.rs
andrei-marinica: `transfer_tokens` is better, yes

catalinnnn: NFTs are ESDTs. transfer_tokens means any tokens, including EGLD. this only works with ESDT, hence the naming

dorin-iancu (approved):

andrei-marinica (approved):

claudiulataretu (approved):

PR #222, opened 2021-05-10
Log2, Sqrt, Pow, elliptic curves functionalities

andrei-diaconescu:


contracts/feature-tests/basic-features/src/lib.rs
```diff
@@ -1227,6 +1258,123 @@ pub trait BasicFeatures {
...
+			size_of_field,
+		);
+		self.crypto().add_ec(
+			&curve,
+			x_first_point,
+			y_first_point,
+			x_second_point,
+			y_second_point,
+		)
+	}
```
alyn509, line 1291: why not create yourself an equation struct and pass just 1 argument to the function in stead of 11
start from here: https://refactoring.guru/introduce-parameter-object

alyn509: the people working with your features will thank you later!

andrei-marinica (approved):

PR #224, opened 2021-05-11
Crowdfunding lottery update

dorin-iancu:
Updated crowdfunding and lottery with up to date framework and mandos features.


claudiu725 (approved):

catalinnnn:
Would rename from "CrowdFunding-esdt" and "Lottery-esdt" to "CrowdFunding" and "Lottery" since now they work with both egld and fungible esdt

PR #257, opened 2021-06-07
github actions - elrond-wasm-node

andrei-marinica:
DOESN'T YET WORK.

Maybe support will be added in the future?
https://www.bountysource.com/teams/actions-rs/issues?tracker_ids=154738033


andrei-marinica:
The problem this was trying to fix was actually solved in https://github.com/ElrondNetwork/elrond-wasm-rs/pull/872

PR #266, opened 2021-06-20
Attributes refactor

andrei-marinica:
Unused attributes now end up in the macro-generated code. This does 2 things:
- (most important) wrong attributes are more obvious. The compiler will complain about any unknown attributes
- it is now possible to annotate endpoints with things like `#[allow(clippy::too_many_arguments)]`


elrond-wasm-derive/src/parse/argument_parse.rs
```diff
@@ -76,3 +35,112 @@ fn missing_self_panic(m: &syn::TraitItemMethod) -> ! {
 		m.sig.ident.to_string()
 	)
 }
+
+fn extract_method_arg(pat_typed: &syn::PatType) -> MethodArgument {
+	let pat = &*pat_typed.pat;
+	let ty = &*pat_typed.ty;
+	let mut arg_metadata = ArgMetadata {
```
dorin-iancu, line 42: Could maybe have a `new()` function for ArgMetadata that creates one with the default values you're using here.

dorin-iancu (approved):

elrond-wasm-derive/src/parse/method_parse.rs
```diff
@@ -1,27 +1,110 @@
...
+) {
+	for attr in attrs {
+		let processed = process_attribute_pass_1(attr, pass_1_data);
+		if !processed {
+			pass_1_unprocessed_attributes.push(attr.clone());
+		}
 	}
 }
+
+fn process_attribute_pass_1(
```
catalinnnn, line 71: Not a fan of this naming model for two reasons: it's not clear what pass_1 does different that pass_2 + is also pretty easy to mix them up, having only one character that differs. At least with a similar naming like pass_first and pass_second you can resolve the second issue, although a more meaningful name would be preffered

catalinnnn (dismissed):

catalinnnn (approved):

dorin-iancu (approved):

PR #269, opened 2021-06-30
big int to small int API

andrei-marinica:


catalinnnn (approved):

elrond-wasm-debug/src/api/big_int_api_mock.rs
```diff
@@ -256,4 +256,35 @@ impl elrond_wasm::api::BigIntApi for RustBigInt {
...
+					}
+				} else {
+					None
+				}
+			},
+			Sign::Minus => {
+				if digits.len() == 1 {
+					let as_u64 = digits[0];
+					match as_u64.cmp(&0x8000000000000000u64) {
+						Ordering::Less => Some(-(as_u64 as i64)),
```
dorin-iancu, line 280: Wonder if there would be a way to use i64::MIN here, but with Rust's type safety, probably not

dorin-iancu (approved):

elrond-wasm-debug/src/api/big_int_api_mock.rs
andrei-marinica: tried it, I cannot write -i64::MIN because it "causes overflow"

andrei-marinica: I decided for some mocks it is good enough

PR #271, opened 2021-07-05
combine some send_api functions + update in contracts

dorin-iancu:
- `self.send().direct()` now takes a nonce as argument. `direct_nft` removed.
- combined `esdt_local_mint` and `esdt_nft_add_quantity`
- combined `esdt_local_burn` and `esdt_nft_burn`

### DIFF
--- a/elrond-wasm/src/api/send_api.rs
+++ b/elrond-wasm/src/api/send_api.rs
@@ -67,41 +67,9 @@ pub trait SendApi: ErrorApi + Clone + Sized {
 		arg_buffer: &ArgBuffer,
 	) -> Result<(), &'static [u8]>;
 
-	/// Performs a simple ESDT transfer, but via async call.
-	/// This is the preferred way to send ESDT.
-	fn transfer_esdt_via_async_call(
-		&self,
-		to: &Address,
-		token: &TokenIdentifier,
-		amount: &Self::AmountType,
-		data: &[u8],
-	) -> ! {
-		let mut serializer = HexCallDataSerializer::new(ESDT_TRANSFER_STRING);
-		serializer.push_argument_bytes(token.as_esdt_identifier());
-		serializer.push_argument_bytes(amount.to_bytes_be().as_slice());
-		if !data.is_empty() {
-			serializer.push_argument_bytes(data);
-		}
-		self.async_call_raw(to, &Self::AmountType::zero(), serializer.as_slice())
-	}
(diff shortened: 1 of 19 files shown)


elrond-wasm/src/api/send_api.rs
```diff
@@ -131,6 +106,46 @@ pub trait SendApi: ErrorApi + Clone + Sized {
...
+	/// Note that EGLD can NOT be transfered with this function.  
+	fn transfer_esdt_via_async_call(
+		&self,
+		to: &Address,
+		token: &TokenIdentifier,
+		nonce: u64,
+		amount: &Self::AmountType,
+		data: &[u8],
+	) -> ! {
+		if nonce == 0 {
```
sasurobert, line 122: can't we combine EGLD transfer here as well ?

elrond-wasm/src/api/send_api.rs
```diff
@@ -131,6 +106,46 @@ pub trait SendApi: ErrorApi + Clone + Sized {
 		)
 	}
 
+	/// Performs a simple ESDT/NFT transfer, but via async call.  
+	/// As with any async call, this immediately terminates the execution of the current call.  
+	/// So only use as the last call in your endpoint.  
+	/// If you want to perform multiple transfers, use `self.send().direct()` instead.  
+	/// Note that EGLD can NOT be transfered with this function.
```
sasurobert, line 113: so right new we have direct and transfer_via_async_call combined - can we do the same for executeOnDestination contract with token transfer ?

elrond-wasm/src/api/send_api.rs
dorin-iancu: Can EGLD be transferred with AsyncCall? Keep in mind this function doesn't call any endpoint, it performs a simple transfer.

elrond-wasm/src/api/send_api.rs
dorin-iancu: executeOnDest context is a bit of a special case. There is a separate way of doing this: Andrei Marinica created a way of importing a contract's proxy, which lets the "parent" contract be able to call endpoints as if they were simple functions. This proxy automatically handles #[payment] arguments and performs the needed contract call.

TL;DR: These functions are not used directly.

catalinnnn (approved):

catalinnnn (approved):

PR #273, opened 2021-07-08
Governance module

dorin-iancu:
Module that allows any smart contract to use governance features. It uses a governance token ESDT for voting and proposing.
- propose actions, by depositing a certain amount of governance tokens
- after a waiting period, the `voting` can start. Users may also `downvote` a proposal.
- once the voting period ends, if the proposal has reached the `quorum`, it may be `queued`, otherwise, it can be `canceled`
- after a waiting period, the `queued` action may be `executed`
- after its execution or cancellation, the voters/downvoters can retrieve their governance tokens


contracts/modules/elrond-wasm-module-governance/src/governance_configurable.rs
```diff
@@ -0,0 +1,197 @@
+elrond_wasm::imports!();
+
+#[elrond_wasm_derive::module]
+pub trait GovernanceConfigurablePropertiesModule {
```
catalinnnn, line 4: I think this module can be merged with the Governance module. Anyone who would want to use the Governance would want to only import one thing

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+		let caller = self.blockchain().get_caller();
+		let governance_token_id = self.governance_token_id().get();
+		let nr_votes_tokens = self.votes(proposal_id).get(&caller).unwrap_or_default();
+		let nr_downvotes_tokens = self.downvotes(proposal_id).get(&caller).unwrap_or_default();
+
+		if nr_votes_tokens > 0 {
+			self.send()
+				.direct(&caller, &governance_token_id, &nr_votes_tokens, &[]);
+
+			self.votes(proposal_id).remove(&caller);
```
catalinnnn, line 43: It's a best practice to clear the storage and then send the tokens. Right now there's no risk of re-entrancy since there's no function call, but still if someone decides to add, he should not worry about this risk

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+		}
+		if nr_downvotes_tokens > 0 {
+			self.send()
```
catalinnnn, line 46: Why not merge with the first send and send only once?

contracts/modules/elrond-wasm-module-governance/src/governance_configurable.rs
```diff
@@ -0,0 +1,197 @@
...
+		require!(
+			caller == sc_address,
+			"Only the SC itself may call this function"
+		);
+
+		Ok(())
+	}
+
+	fn try_change_quorum(&self, new_value: Self::BigUint) -> SCResult<()> {
+		require!(new_value > 0, "Quorum can't be set to 0");
```
catalinnnn, line 113: Consider using != 0 instead of > 0.

contracts/modules/elrond-wasm-module-governance/src/governance_configurable.rs
```diff
@@ -0,0 +1,197 @@
...
+		self.try_change_voting_period_in_blocks(voting_period_in_blocks)?;
+		self.try_change_lock_time_after_voting_ends_in_blocks(
+			lock_time_after_voting_ends_in_blocks,
+		)?;
+
+		Ok(())
+	}
+
+	// endpoints - these can only be called by the SC itself.
+	// i.e. only by proposing and executing an action with the SC as dest and the respectiv func name
```
catalinnnn, line 42: Typo respective

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+
+	#[payable("*")]
+	#[endpoint]
+	fn propose(
+		&self,
+		#[payment_amount] payment_amount: Self::BigUint,
+		description: BoxedBytes,
+		#[var_args] actions: VarArgs<GovernanceActionAsMultiArg<Self::BigUint>>,
+	) -> SCResult<usize> {
+		self.require_payment_token_governance_token()?;
```
catalinnnn, line 63: For consistency I think it's best that you either take payment info from #payment parameters or for call value and not from both

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+	#[view(getProposalDescription)]
+	fn get_proposal_description(&self, proposal_id: usize) -> OptionalArg<BoxedBytes> {
+		if !self.proposal_exists(proposal_id) {
+			OptionalArg::None
+		} else {
+			OptionalArg::Some(self.proposals().get(proposal_id).description)
+		}
+	}
+
+	#[view]
```
catalinnnn, line 307: Why no camel case here?

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+	fn get_proposal_actions(
+		&self,
+		proposal_id: usize,
+	) -> MultiResultVec<GovernanceActionAsMultiArg<Self::BigUint>> {
+		if !self.proposal_exists(proposal_id) {
```
catalinnnn, line 312: Replace with require

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+		require!(
+			current_block >= lock_end,
+			"Proposal is in timelock status. Try again later"
+		);
+
+		let proposal = self.proposals().get(proposal_id);
+		let total_gas_needed = self.total_gas_needed(&proposal.actions);
+		let gas_left = self.blockchain().get_gas_left();
+
+		// reserve double to account for the processing needed for each action, like argument serialization
```
catalinnnn, line 195: I'd expect total gas needed to be the total gas needed, not the total gas needed just for action (not including argument serialization). Also remove the comment and try to express the intention with another variable. 
gas_needed_for_action = ...
gas_needed_for_serialization = gas_needed_for_action
total_gas_needed = gas_needed_for_action + gas_needed_for_serialization

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+pub mod governance_configurable;
+
+pub mod governance_proposal;
+use governance_proposal::*;
+
+#[elrond_wasm_derive::module]
+pub trait GovernanceModule:
+	governance_configurable::GovernanceConfigurablePropertiesModule
+{
+	// TODO: Withdraw functionality after voting ends
```
catalinnnn, line 14: Remove TODO or implement withdraw

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+
+	#[storage_mapper("governance:downvotes")]
+	fn downvotes(&self, proposal_id: usize) -> MapMapper<Self::Storage, Address, Self::BigUint>;
+
+	/// Could be calculated by iterating over the "votes" mapper, but that costs a lot of gas
+	#[view(getTotalVotes)]
+	#[storage_mapper("governance:totalVotes")]
+	fn total_votes(&self, proposal_id: usize) -> SingleValueMapper<Self::Storage, Self::BigUint>;
+
+	/// Could be calculated by iterating over the "downvotes" mapper, but that costs a lot of gas
```
catalinnnn, line 404: Not a fan of this comments. If someone changes downvotes mapper he will 99% forget to change this comment also

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+	fn total_gas_needed(&self, actions: &[GovernanceAction<Self::BigUint>]) -> u64 {
+		let mut total = 0;
+		for action in actions {
+			total += action.gas_limit;
+		}
+
+		total
+	}
+
+	fn get_sc_balance(&self, token_id: &TokenIdentifier) -> Self::BigUint {
```
catalinnnn, line 361: Unused function

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+	fn is_valid_proposal_id(&self, proposal_id: usize) -> bool {
+		proposal_id >= 1 && proposal_id <= self.proposals().len()
+	}
+
+	fn proposal_exists(&self, proposal_id: usize) -> bool {
+		self.is_valid_proposal_id(proposal_id) && !self.proposals().item_is_empty(proposal_id)
+	}
+
+	fn total_gas_needed(&self, actions: &[GovernanceAction<Self::BigUint>]) -> u64 {
+		let mut total = 0;
```
catalinnnn, line 353: Check for overflow maybe?

CHANGELOG.md
```diff
@@ -12,7 +12,7 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 ## [elrond-wasm 0.17.4] - 2021-06-30
 - conversions from big ints to small int: `BigUint::to_u64`, `BigInt::to_i64`
 
-## [elrond-wasm 0.17.3] - 2021-06-11
+## [elrond-wasm 0.17.4] - 2021-06-11
```
catalinnnn, line 15: Why change a past changelog? Should add another entry for next version 0.17.5

contracts/modules/elrond-wasm-module-governance/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+
+	/// Not stored under "proposals", as that would require deserializing the whole struct
+	#[storage_mapper("governance:proposalStartBlock")]
+	fn proposal_start_block(&self, proposal_id: usize) -> SingleValueMapper<Self::Storage, u64>;
+
+	#[storage_mapper("governance:proposalQueueBlock")]
+	fn proposal_queue_block(&self, proposal_id: usize) -> SingleValueMapper<Self::Storage, u64>;
+
+	#[storage_mapper("governance:votes")]
+	fn votes(&self, proposal_id: usize) -> MapMapper<Self::Storage, Address, Self::BigUint>;
```
catalinnnn, line 394: Consider renaming to upvotes

contracts/modules/elrond-wasm-module-governance/src/governance_configurable.rs
dorin-iancu: That's true, but still, even if it takes a slightly more complicated setup (read: add two modules instead of one to the main SC), the source code is a lot cleaner to read.

dorin-iancu: Would be nice if including the "main" module would automatically include the sub-modules as well, but unfortunately, that's not currently possible, so I went for this compromise.

contracts/modules/elrond-wasm-module-governance/src/governance_configurable.rs
dorin-iancu: Fixed.

contracts/modules/elrond-wasm-module-governance/src/governance_configurable.rs
dorin-iancu: Changed everywhere in this file.

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Already implemented. Comment removed.

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Fixed. Also realized the `remove()` method checks for empty, so there is no need to manually check for empty.

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Fixed.

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Aside from the generic deposit function, all payable functions in this module take only governance tokens as payment. In any other case, I'd change, but in this specific situation, I believe this is fine.

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Fair point. Fixed.

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Forgot :)
Fixed.

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Cannot use `require!`, as this view function does not return `SCResult<T>` (which is intended, I prefer returning "empty" instead of error for view functions)

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Removed.

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Added check for each action to have less than max gas limit per block (1.5 billion), so this can never overflow.

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Fair point, but I currently don't have a better solution. Really don't want to iterate over the whole mapper at every vote/downvote.

contracts/modules/elrond-wasm-module-governance/src/lib.rs
dorin-iancu: Actually, removed the "double" restriction. Will let the user handle the gas costs instead of trying to estimate in the SC. Only restriction is to have more gas than the total the actions need.

CHANGELOG.md
dorin-iancu: Mass replaced, didn't notice I also changed it here. Fixed.

catalinnnn (approved):

catalinnnn (approved):

PR #278, opened 2021-07-15
Random number generator module

dorin-iancu:


contracts/modules/elrond-wasm-module-rng/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	/// please use one of the specific methods instead of directly calling this
+	fn _next_number(&self, nr_bytes: usize) -> u64 {
+		if nr_bytes == 0 || nr_bytes as u32 > u64::BITS / u8::BITS {
+			return 0;
+		}
+
+		let mut seed = self.rng_seed().get();
+		let mut start = self.rng_seed_index().get();
```
catalinnnn, line 56: Does start have to be mut?

contracts/modules/elrond-wasm-module-rng/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+		let end = start + nr_bytes;
+		self.rng_seed_index().set(&end);
+
+		u64::top_decode(&seed[start..end]).unwrap_or_default()
+	}
+
+	/// Parent contract should never call this function. Calling this directly leads to undefined behaviour.
+	/// Add new seed to old seed for cases when many random numbers are needed in the same block
+	/// This prevents infinite loops, where a contract may want distinct numbers (for example, lottery SC)
+	fn _refresh_seed(&self, old_seed: &mut RngSeed) {
```
catalinnnn, line 72: Would pub(crate) work here?

contracts/modules/elrond-wasm-module-rng/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	fn next_u64(&self) -> u64 {
+		let nr_bytes = (u64::BITS / u8::BITS) as usize;
+		let rand_nr = self._next_number(nr_bytes);
+
+		// since wasm32 doesn't have u64, but only i64, we set the first bit to 0
+		// this is done to prevent any potential unintended bugs that are impossible to track down
+	
+		rand_nr & 0x7fffffffffffffff
```
catalinnnn, line 46: I think having a random of u64 always starting with 0 is not a good idea. I'd suggest creating a new endpoint like next_i64 and using next_u64 & 0x7fffffffffffffff into it. This way you can also remove the comment in the code

contracts/modules/elrond-wasm-module-rng/src/lib.rs
```diff
@@ -0,0 +1,90 @@
+#![no_std]
+
+elrond_wasm::imports!();
+
+const SEED_SIZE: usize = 48;
+type RngSeed = Box<[u8; SEED_SIZE]>;
+
+/// The module deals with generating random numbers
```
catalinnnn, line 8: Too obvious comment. I'd suggest removing it

contracts/modules/elrond-wasm-module-rng/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+		}
+
+		self.rng_seed().set(old_seed);
+		self.rng_seed_index().clear();
+	}
+
+	// storage
+
+	#[storage_mapper("rng:seed_index")]
+	fn rng_seed_index(&self) -> SingleValueMapper<Self::Storage, usize>;
```
catalinnnn, line 86: Why do you need rng_seed_index?

contracts/modules/elrond-wasm-module-rng/src/lib.rs
dorin-iancu: Yes, otherwise it can't be modified in the following `if` statement.

contracts/modules/elrond-wasm-module-rng/src/lib.rs
dorin-iancu: pub(crate) only works for structs. Traits cannot have private functions.

contracts/modules/elrond-wasm-module-rng/src/lib.rs
dorin-iancu: So I know where I left off in the seed, i.e. determining that start..end range. Otherwise, I'd reuse the same numbers over and over.

catalinnnn (approved):

PR #285, opened 2021-07-21
Only owner endpoint

alyn509:


elrond-wasm/src/api/blockchain_api.rs
```diff
@@ -18,6 +18,12 @@ pub trait BlockchainApi: StorageReadApi + ErrorApi + Clone + Sized + 'static {
 
 	fn get_owner_address(&self) -> Address;
 
+	fn check_caller_is_owner(&self, endpoint: &'static str) {
```
catalinnnn, line 21: Please consider changing to "endpoint" to "error_message"

elrond-wasm-derive/src/generate/only_owner_gen.rs
```diff
@@ -0,0 +1,16 @@
+use crate::model::{Method, PublicRole};
+
+pub fn generate_only_owner_snippet(m: &Method) -> proc_macro2::TokenStream {
+	if let PublicRole::Endpoint(endpoint_metadata) = &m.public_role {
+		if endpoint_metadata.only_owner {
+			let message = format!(
+				"Only owner can call {} !",
```
catalinnnn, line 7: The "!" may seem too verbose. Please consider removing the "!" sign

contracts/feature-tests/basic-features/mandos/only_owner.scen.json
```diff
@@ -0,0 +1,96 @@
...
+                "address:an_account": {
+                    "nonce": "1",
+                    "balance": "0",
+                    "storage": {},
+                    "code": ""
+                }
+            }
+        }
+    ]
+}
```
catalinnnn, line 96: Add empty line

elrond-wasm-derive/src/generate/only_owner_gen.rs
dorin-iancu: There is no need to have the endpoint name in the error message, as you already know the endpoint you called. Would be more efficient if all had the same error message, as that would just have a single string message constant in the output code instead of one message for each owner-only endpoint.

contracts/feature-tests/basic-features/src/macros.rs
```diff
@@ -11,6 +11,12 @@ pub trait Macros {
 		Ok(())
 	}
 
+	#[only_owner]
+	#[endpoint]
+	fn only_owner_endpoint(&self) -> SCResult<()> {
```
andrei-marinica, line 16: No need for SCResult here, the annotation should work regardless.

andrei-marinica (changes requested):

catalinnnn (dismissed):

elrond-wasm/src/api/blockchain_api.rs
```diff
@@ -18,6 +18,12 @@ pub trait BlockchainApi: StorageReadApi + ErrorApi + Clone + Sized + 'static {
 
 	fn get_owner_address(&self) -> Address;
 
+	fn check_caller_is_owner(&self) {
+		if self.get_owner_address() != self.get_caller() {
+			self.signal_error(b"Only owner can call this");
```
andrei-marinica, line 23: Endpoint can only be called by owner
parcă e mai bine

andrei-marinica (changes requested):

dorin-iancu (approved):

dorin-iancu (approved):

andrei-marinica (approved):

PR #290, opened 2021-07-27
Modernize examples

dorin-iancu:
- use #[only_owner] annotation where possible
- use SingleValueMapper instead of storage_set/storage_get pairs (also use update() where possible instead of get, modify, set)
- use proxy instead of manually serializing async call data
- use API get_owner function instead of manually keeping the owner in storage


contracts/examples/crypto-kitties/kitty-auction/src/lib.rs
```diff
@@ -314,7 +312,7 @@ pub trait KittyAuction {
 		}
 	}
 
-	fn _get_kitty_ownership_contract_address_or_default(&self) -> Address {
+	fn get_kitty_ownership_contract_address_or_default(&self) -> Address {
 		if self.is_empty_kitty_ownership_contract_address() {
```
catalinnnn, line 316: This looks like an old style mapper also. Is it the case?

PR #298, opened 2021-07-30
Managed buffers take 1 [ABANDONED]

andrei-marinica:
This branch has reached a dead end, will probably be abandoned.


andrei-marinica:
Approach to managed types permanently abandoned.

PR #311, opened 2021-08-18
Token Attributes Mapper

catalinnnn:


elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
```diff
@@ -0,0 +1,239 @@
...
+        );
+    }
+
+    fn is_empty_mapping_value(&self, token_id: &TokenIdentifier) -> bool {
+        self.api
+            .storage_load_len(self.build_key_token_id_mapping(token_id).as_slice())
+            == 0
+    }
+
+    fn get_token_attributes_value(&self, mapping: u8, token_nonce: u64) -> Vec<u8> {
```
andrei-marinica, line 207: Make this generic.
get_token_attributes_value<T: TopEncode + TopDecode>

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
```diff
@@ -0,0 +1,239 @@
...
+        storage_get(
+            self.api.clone(),
+            self.build_key_token_attr_value(mapping, token_nonce)
+                .as_slice(),
+        )
+    }
+
+    fn set_token_attributes_value(&self, mapping: u8, token_nonce: u64, value: Vec<u8>) {
```
andrei-marinica, line 215: Make this generic.
set_token_attributes_value<T: TopEncode + TopDecode>

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
```diff
@@ -0,0 +1,239 @@
...
+
+        let mapping = self.get_mapping_value(token_id);
+        let has_value = !self.is_empty_token_attributes_value(mapping, token_nonce);
+        if !has_value {
+            self.api
+                .signal_error(VALUE_NOT_PREVIOUSLY_SET_ERROR_MESSAGE);
+        }
+
+        let value = self.get_token_attributes_value(mapping, token_nonce);
+        T::top_decode(value).unwrap()
```
andrei-marinica, line 142: Don't manually decode and avoid `unwrap` (since it panics).
Use the power of generics to extract the right type directly from `storage_get`.
`storage_get` will also handle errors for you, plus it knows how to handle managed types, which `Vec<u8>` does not.

andrei-marinica (changes requested):

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
```diff
@@ -0,0 +1,239 @@
...
+            counter
+        };
+
+        let has_value = !self.is_empty_token_attributes_value(mapping, token_nonce);
+        if has_value {
+            self.api.signal_error(VALUE_ALREADY_SET_ERROR_MESSAGE);
+        }
+
+        let mut value = Vec::<u8>::new();
+        attributes.top_encode(&mut value).unwrap();
```
dorin-iancu, line 73: Very dangerous to assume the top encoding can never return errors. This might be fine for most cases with derived implementations of TopEncode, but keep in mind users can have custom implementations, which can return errors.

I think the best approach would be something like 
match attributes.top_encode(&mut value) {
Result::Ok(_) => { self.set_token_attributes_value(mapping, token_nonce, value); }
Result::Err(_) => *signal error here*

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
catalinnnn: The idea was to have a storage for all types of custom Attrs struct

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
```diff
@@ -0,0 +1,239 @@
...
+                self.api.signal_error(COUNTER_OVERFLOW_ERROR_MESSAGE);
+            }
+
+            counter += 1;
+            self.set_mapping_value(token_id, counter);
+            self.set_counter_value(counter);
+            counter
+        };
+
+        let has_value = !self.is_empty_token_attributes_value(mapping, token_nonce);
```
dorin-iancu, line 67: Since this seems to be repeated in other functions as well, it would be best to have a separate function the "has_value" check instead of manually checking for is_empty.

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
```diff
@@ -0,0 +1,239 @@
...
+
+        let mapping = self.get_mapping_value(token_id);
+        let has_value = !self.is_empty_token_attributes_value(mapping, token_nonce);
+        if !has_value {
+            self.api
+                .signal_error(VALUE_NOT_PREVIOUSLY_SET_ERROR_MESSAGE);
+        }
+
+        let mut value = Vec::<u8>::new();
+        attributes.top_encode(&mut value).unwrap();
```
dorin-iancu, line 97: Same comment as above regarding unwrap(). Might be better to have a separate function that only handles attributes encoding.

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub fn clear(&self, token_id: &TokenIdentifier, token_nonce: u64) {
+        let has_mapping = !self.is_empty_mapping_value(token_id);
+        if !has_mapping {
+            self.api.signal_error(UNKNOWN_TOKEN_ID_ERROR_MESSAGE);
+        }
+
+        let mapping = self.get_mapping_value(token_id);
+        let has_value = !self.is_empty_token_attributes_value(mapping, token_nonce);
+        if !has_value {
+            self.api
```
dorin-iancu, line 110: I think we should simply return here instead of signaling an error. That would be consistent with other mappers implementation as well.

Could be simply re-written as
if has_value {
self.clear_token_attributes_value(mapping, token_nonce);
}

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
```diff
@@ -0,0 +1,239 @@
...
+                .signal_error(VALUE_NOT_PREVIOUSLY_SET_ERROR_MESSAGE);
+        }
+
+        self.clear_token_attributes_value(mapping, token_nonce);
+    }
+
+    pub fn is_empty(&self, token_id: &TokenIdentifier, token_nonce: u64) -> bool {
+        let has_mapping = self.is_empty_mapping_value(token_id);
```
dorin-iancu, line 118: This seems wrong. Shouldn't has_mapping be the opposite of is_empty?

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
```diff
@@ -0,0 +1,239 @@
...
+        let mapping = self.get_mapping_value(token_id);
```
dorin-iancu, line 119: Shouldn't you check has_mapping before getting the mapping?

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
dorin-iancu, line 142: no unchecked unwrapping please

dorin-iancu (changes requested):

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
catalinnnn: What's the difference? Signal error gets called every time

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
catalinnnn: Thank you. Was searching for this for some minutes :)

dorin-iancu (approved):

andrei-marinica (approved):

elrond-wasm/src/storage/mappers/token_attributes_mapper.rs
```diff
@@ -0,0 +1,239 @@
...
+        attributes: &T,
+    ) {
+        let has_mapping = !self.is_empty_mapping_value(token_id);
+
+        let mapping = if has_mapping {
+            self.get_mapping_value(token_id)
+        } else {
+            let mut counter = self.get_counter_value();
+            if counter == u8::MAX {
+                self.api.signal_error(COUNTER_OVERFLOW_ERROR_MESSAGE);
```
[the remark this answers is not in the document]
andrei-marinica: Yeah, it could say `256` directly

PR #320, opened 2021-08-25
Multi transfer exec on dest

catalinnnn:


elrond-wasm/src/types/interaction/contract_call.rs
```diff
@@ -46,9 +52,13 @@ where
     SA: SendApi + 'static,
 {
     let mut contract_call = ContractCall::<SA, R>::new(api, to, endpoint_name);
-    contract_call.payment_token = payment_token;
-    contract_call.payment_amount = payment_amount;
-    contract_call.payment_nonce = payment_nonce;
+    let payment = EsdtTokenPayment {
+        token_type: EsdtTokenType::Invalid,
```
dorin-iancu, line 56: Might not be such a great idea to use `EsdtTokenType::Invalid` here, even though currently this field isn't really used. I think the best approach would be to add a static method in `EsdtTokenPayment`, something like `from_fields(token_id, nonce, amount) -> Self`, which automatically sets `token_type` accordingly and returns a new instance.

elrond-wasm/src/types/interaction/contract_call.rs
```diff
@@ -75,13 +83,25 @@ where
...
+        self.payments[0].token_nonce = payment_nonce;
+        self
+    }
+
+    pub fn with_multi_token_transfer(
+        mut self,
+        payments: Vec<EsdtTokenPayment<SA::AmountType>>,
+    ) -> Self {
+        if payments.is_empty() {
+            self.api.signal_error(b"Payment vec should not be empty");
```
dorin-iancu, line 101: I'm not sure signaling an error here is the best approach. Maybe just `return self` instead?
```
if !payments.is_empty() {
self.payment = payments;
}

self
```

elrond-wasm/src/types/interaction/contract_call.rs
```diff
@@ -102,21 +122,27 @@ where
...
     fn convert_to_esdt_transfer_call(self) -> Self {
-        if self.payment_token.is_egld() {
+        if self.payments.len() == 1 {
+            self.convert_to_single_transfer_esdt_call()
+        } else {
+            self.convert_to_multi_transfer_esdt_call()
+        }
+    }
+
+    fn convert_to_single_transfer_esdt_call(self) -> Self {
```
dorin-iancu, line 132: Just for readability purposes, I think it would be better to have something like `let payment = &self.payments[0];` and use that instead of indexing everywhere. Way less prone to mistakes too.

elrond-wasm/src/types/interaction/contract_call.rs
```diff
@@ -270,21 +317,29 @@ where
...
+            self.single_transfer_execute()
+        } else {
+            self.multi_transfer_execute()
+        }
+    }
+
+    fn single_transfer_execute(&self) {
         let gas_limit = self.resolve_gas_limit_with_leftover();
-        let result = if self.payment_token.is_egld() {
+        let result = if self.payments[0].token_name.is_egld() {
```
dorin-iancu, line 329: Same "reference to indexing" suggestion as above

dorin-iancu (changes requested):

dorin-iancu (dismissed):

PR #325, opened 2021-08-28
ManagedVec

andrei-marinica:
A structure based on a managed buffer, that can hold meaningful content.
Can be used to create "managed list of managed buffers" or of other managed types.
Highly extensible.

The idea is to use a managed buffer containing handles to other managed buffers to send log topics to Arwen, or to pass arguments. Also to retrieve results from execute_on_dest_context.

### DIFF
--- a/elrond-wasm/src/types/managed/managed_vec_item.rs
+++ b/elrond-wasm/src/types/managed/managed_vec_item.rs
@@ -7,7 +7,11 @@ use crate::{
 
 use super::{BigInt, BigUint, ManagedBuffer, ManagedType};
 
-/// Types that implement this trait can fir inside a `ManagedVec`.
+/// Types that implement this trait can be items inside a `ManagedVec`.
+/// All these types need a payload, i.e a representation that gets stored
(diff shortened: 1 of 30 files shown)


elrond-wasm/src/types/managed/managed_vec_item.rs
```diff
@@ -0,0 +1,69 @@
+use elrond_codec::{NestedDecode, NestedEncode};
+
+use crate::{
+    abi::TypeAbi,
+    api::{Handle, ManagedTypeApi},
+};
+
+use super::{BigInt, BigUint, ManagedBuffer, ManagedType};
+
+/// Types that implement this trait can fir inside a `ManagedVec`.
```
dorin-iancu, line 10: fir -> fit, though I'm not entirely sure the wording is correct even so. Maybe `Types that implement this trait can be stored inside a ManagedVec`?

dorin-iancu (approved):

dorin-iancu (approved):

PR #334, opened 2021-09-07
nft-minter and seller module

dorin-iancu:


contracts/examples/nft-minter/src/nft_module.rs
```diff
@@ -0,0 +1,220 @@
...
+        require!(royalties <= ROYALTIES_MAX, "Royalties cannot exceed 100%");
+
+        let nft_token_id = self.nft_token_id().get();
+
+        let mut serialized_attributes = Vec::new();
+        attributes.top_encode(&mut serialized_attributes)?;
+
+        let attributes_hash = self.crypto().sha256(&serialized_attributes);
+
+        let nft_nonce = self.send().esdt_nft_create(
```
sasurobert, line 174: so this is for PURE NFTs ? always value == 1 ? Can we extend for sft's as well?

contracts/examples/nft-minter/src/nft_module.rs
```diff
@@ -0,0 +1,220 @@
...
+            &nft_token_id,
+            &self.types().big_uint_from(NFT_AMOUNT),
```
sasurobert, line 176: if we extend to SFTs we might combine with the NONCE-Attributes mapper.

sasurobert: and like have somewhere a switch which determines if we use nonce-attributes mapper or not.

contracts/examples/nft-minter/src/nft_module.rs
dorin-iancu: It could, but I wanted this to be as simple and as specific as possible. I'd have to add a lot of conditions otherwise, which would only make it more confusing.

dorin-iancu: For instance, I'd have to have separate cases for issue/set roles functions.

claudiu725 (approved):

PR #343, opened 2021-09-10
Real linked list mapper

catalinnnn:
Renamed old LinkedListMapper to QueueMapper
Added a new LinkedListMapper with proper delete/insert and iterate


elrond-wasm/src/storage/mappers/linked_list_mapper.rs
```diff
@@ -155,197 +181,245 @@ where
...
+        if self.is_empty_node(node.node_id) {
+            return None;
+        }
+
+        let mut info = self.get_info();
+        let new_node_id = info.generate_new_node_id();
+
+        let new_node_prev_id = node.prev_id;
+        node.prev_id = new_node_id;
+        self.set_node(node.node_id, &node);
```
dorin-iancu, line 263: Same as above
```suggestion
        self.set_node(node.node_id, node);
```

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
```diff
@@ -155,197 +181,245 @@ where
...
-    /// Provides a forward iterator.
+    pub fn remove_node_by_id(&mut self, node_id: u32) -> Option<T> {
+        if self.is_empty_node(node_id) {
+            return None;
+        }
+
+        let node = self.get_node(node_id);
+        self.remove_node(&node);
+        Some(node.value)
+    }
```
dorin-iancu, line 399: For consistency with the `get_node_by_id` method, this should also return the full node.

```suggestion
    pub fn remove_node_by_id(&mut self, node_id: u32) -> Option<LinkedListNode<T>> {
        if self.is_empty_node(node_id) {
            return None;
        }

        let node = self.get_node(node_id);
        self.remove_node(&node);
        Some(node)
    }
```

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
```diff
@@ -9,36 +9,58 @@ use crate::{
...
+    pub(crate) value: T,
+    pub(crate) node_id: u32,
+    pub(crate) next_id: u32,
+    pub(crate) prev_id: u32,
+}
+
+impl<T: NestedEncode + NestedDecode + TopEncode + TopDecode + Clone> LinkedListNode<T> {
+    pub fn get_value(&self) -> T {
+        self.value.clone()
+    }
```
dorin-iancu, line 34: Most of the time, developers won't care about any field other than the `value`, so cloning on every `get` is an unneeded performance penalty, especially if we're dealing with managed types.

Keep this method as it is (aside from renaming), but also add another two (note: not 100% sure about the syntax here, please check if it compiles)
```suggestion
    pub fn get_value_cloned(&self) -> T {
        self.value.clone()
    }
    
    pub fn get_value_as_ref<'a>(&'a self) -> &'a T {
        &self.value
    }
    
    pub fn into_value(self) -> T {
        self.value
    }
    
```

elrond-wasm/src/storage/mappers/linked_list_mapper.rs
```diff
@@ -423,39 +487,48 @@ where
...
         if current_node_id == NULL_ENTRY {
             return None;
         }
-        self.node_id = self.linked_list.get_node(current_node_id).next;
-        Some(self.linked_list.get_value(current_node_id))
+
+        let node = self.linked_list.get_node(current_node_id);
+        self.node_id = node.next_id;
+        Some(node.value)
     }
```
dorin-iancu, line 525: This method should probably return the whole node now, otherwise it would be difficult to save the current node_id, for instance. (unless I'm missing something)

dorin-iancu (changes requested):

dorin-iancu (dismissed):

PR #347, opened 2021-09-15
Managed multiresultvec

andrei-marinica:
Implemented:
ManagedVarArgs
ManagedMultiResultVec
(they are the same type)


elrond-wasm/src/types/managed/managed_multi_result_vec.rs
```diff
@@ -0,0 +1,168 @@
...
+    T: ManagedVecItem<M> + TopDecode,
+{
+    pub fn to_vec(&self) -> ManagedVec<M, T> {
+        let mut result = ManagedVec::new_empty(self.raw_buffers.type_manager());
+        let serializer = ManagedSerializer::new(self.raw_buffers.type_manager());
+        for item in self.raw_buffers.into_iter() {
+            result.push(serializer.top_decode_from_managed_buffer(&item));
+        }
+        result
+    }
```
dorin-iancu, line 92: For consistency, methods with the naming `to_other_type` should consume `self`:
```suggestion
    pub fn to_vec(self) -> ManagedVec<M, T> {
        let mut result = ManagedVec::new_empty(self.raw_buffers.type_manager());
        let serializer = ManagedSerializer::new(self.raw_buffers.type_manager());
        for item in self.raw_buffers.into_iter() {
            result.push(serializer.top_decode_from_managed_buffer(&item));
        }
        result
    }
```

Maybe also add some rust-docs explaining that this is an expensive operation. It's not immediately obvious, as in the previous version it was a NOP.

mandos/src/esdt.rs
```diff
@@ -0,0 +1,42 @@
+use super::*;
+use std::collections::BTreeMap;
+
+#[derive(Debug)]
+pub struct Esdt {
+    pub token_identifier: Option<BytesValue>,
```
dorin-iancu, line 6: Should this be an Option?

dorin-iancu (approved):

elrond-wasm/src/types/managed/managed_multi_result_vec.rs
andrei-marinica: "into_..." consumes type,, "to_..." does not

mandos/src/esdt.rs
andrei-marinica: I think Alin committed this here by accident. It's ok, we'll return on this.

PR #348, opened 2021-09-16
Nested mapper

claudiu725:
Allow nested storage mappers.
- add methods for manipulating nested storage values:
    - for insertion / removal: `push_nested`, `insert_nested`, `remove_nested`
    - `clear` now works for both plain values and nested storage values
    - add `entry_nested`
    - add nested iterators:`iter_nested`, `values_nested`
- add `IntoStorageMapper` trait which defines `StorageMapperType`


andrei-marinica:
Implementation will be replaced by a new storage mapper mechanism in the near future.

PR #384, opened 2021-10-14
Remove `#[var_args]` annotation

claudiu725:
- work-in-progress


andrei-marinica:
Fixed in https://github.com/ElrondNetwork/elrond-wasm-rs/pull/696

PR #422, opened 2021-10-28
multisig extension

alyn509:


andrei-marinica (approved):

andrei-marinica (approved):

contracts/examples/multisig/mandos/deployOtherMultisig.scen.json
```diff
@@ -0,0 +1,144 @@
...
+                "function": "proposeSCDeployFromSource",
+                "arguments": [
+                    "0",
+                    "sc:multisig",
+                    "false",
+                    "false",
+                    "false",
+                    "1",
+                    "address:paul"
+                ],
```
claudiu725, line 35: The contract having its deploy proposed here is also used when upgrading in the `upgrade.scen.json` test.
As such, it should be upgradable:
```suggestion
                "function": "proposeSCDeployFromSource",
                "arguments": [
                    "0",
                    "sc:multisig",
                    "true",
                    "false",
                    "false",
                    "1",
                    "address:paul"
                ],
```

contracts/examples/multisig/mandos/upgrade.scen.json
```diff
@@ -0,0 +1,166 @@
...
+                "function": "proposeSCUpgrade",
+                "arguments": [
+                    "sc:multisig-child",
+                    "0",
+                    "file:../test-contracts/adder.wasm",
+                    "false",
+                    "false",
+                    "false",
+                    "1234"
+                ],
```
claudiu725, line 24: Similar situation here. The contract which is upgraded here is then used in `upgrade_from_source.scen.json`.
It should be upgradable:
```suggestion
                "function": "proposeSCUpgrade",
                "arguments": [
                    "sc:multisig-child",
                    "0",
                    "file:../test-contracts/adder.wasm",
                    "true",
                    "false",
                    "false",
                    "1234"
                ],
```

elrond-wasm/src/types/managed/managed_vec.rs
```diff
@@ -149,6 +149,13 @@ where
         });
     }
 
+    /// New `ManagedVec` instance with 1 element in it.
+    pub fn singleton(api: M, item: T) -> Self {
```
claudiu725, line 153: I think the name `singleton` is misleading, since this word usually describes the singleton design pattern.
I would suggest a different name for this function (`from_item` / `from_single_item` or something along those lines).

claudiu725 (changes requested):
Assuming the tests were committed as passing, this might mean that the `upgradable` flag is not checked in mandos. Is this the case?

alyn509:
> Assuming the tests were committed as passing, this might mean that the `upgradable` flag is not checked in mandos. Is this the case?

yes

claudiu725 (dismissed):

claudiu725 (approved):

PR #433, opened 2021-11-03
Unordered set mapper

alyn509:
This is basically a SetMapper, but implemented on VecMapper rather than QueueMapper

The main diference is that remove is done via swap_remove: last element gets the index of the one we want to remove and we remove the last index. 

This allows us to easier remove items from greater VecMapper without the requirement of iterating, keeping the cost growth to O(1)


elrond-wasm-debug/tests/test_hash_unordered_set_mapper.rs
```diff
@@ -0,0 +1,82 @@
...
+use elrond_wasm_debug::DebugApi;
+
+fn create_set() -> UnorderedSetMapper<DebugApi, u64> {
+    let api = DebugApi::dummy();
+    let base_key = StorageKey::new(api.clone(), &b"my_unordered_set"[..]);
+    UnorderedSetMapper::new(api, base_key)
+}
+
+fn check_set(set: &UnorderedSetMapper<DebugApi, u64>, expected: Vec<u64>) {
+    println! { "Da lengths: want {:?} have {:?}",set.len(), expected.len()}
```
andrei-marinica: I think we can remove this print

elrond-wasm-debug/tests/test_hash_unordered_set_mapper.rs
```diff
@@ -0,0 +1,82 @@
...
+    assert_eq!(set.len(), expected.len());
+    let actual: Vec<u64> = set.iter().collect();
+    println! { "Da vectors: want {:?} have {:?}",actual, expected}
```
andrei-marinica: ... and this too

andrei-marinica (approved):

PR #434, opened 2021-11-04
swap_remove for VecMapper

alyn509:


elrond-wasm/src/storage/mappers/vec_mapper.rs
```diff
@@ -181,6 +181,31 @@ where
         storage_clear(self.api.clone(), &self.item_key(index));
     }
 
+    /// Clears item at index from storage by swap remove
+    /// last item takes the index of the item to remove
+    /// and we remove the last index.
+    pub fn swap_remove(&mut self, index: usize) -> bool {
+        let _ = self.swap_remove_and_get_old_last(index);
+        true
+    }
```
dorin-iancu, line 190: I don't think we need to return `true` here?
```suggestion
    pub fn swap_remove(&mut self, index: usize) {
        let _ = self.swap_remove_and_get_old_last(index);
    }
```

elrond-wasm/src/storage/mappers/vec_mapper.rs
```diff
@@ -181,6 +181,31 @@ where
...
+
+    pub(crate) fn swap_remove_and_get_old_last(&mut self, index: usize) -> Option<T> {
+        if index == 0 || index > self.len() {
+            self.api.signal_error(&b"index out of range"[..]);
+        }
+
+        let length = self.len();
```
dorin-iancu, line 197: Getting the length twice, extract `self.len()` into the `length` variable before the if statement. Also, for more clarity, I suggest renaming it to `last_item_index` instead of `length`

dorin-iancu (changes requested):

andrei-marinica (dismissed):

dorin-iancu (approved):

PR #438, opened 2021-11-05
Empty callback based on ABI

andrei-marinica:


elrond-wasm-derive/src/generate/abi_gen.rs
```diff
@@ -89,14 +87,28 @@ fn generate_abi_method_body(
...
+        .iter()
+        .filter(|m| {
+            matches!(
+                m.public_role,
+                PublicRole::Callback(_) | PublicRole::CallbackRaw
+            )
+        })
+        .count();
+    callback_count > 0
+}
```
claudiu725, line 108:
```suggestion
fn has_callback(contract: &ContractTrait) -> bool {
    contract
        .methods
        .iter()
        .any(|m| {
            matches!(
                m.public_role,
                PublicRole::Callback(_) | PublicRole::CallbackRaw
            )
        })
}
```
Use `any` instead of `filter` + `count() > 0`.

claudiu725 (changes requested):

claudiu725 (approved):

PR #443, opened 2021-11-09
Cleanup allocator usage from basic operations

andrei-marinica:


andrei-marinica:
Features already implemented elsewhere.

PR #449, opened 2021-11-10
Derive ManagedVecItem

andrei-marinica:


elrond-wasm/src/types/managed/managed_vec_item.rs
```diff
@@ -46,13 +46,16 @@ macro_rules! impl_int {
         }
     };
 }
-
+impl_int! {u8, 1}
+impl_int! {u16, 2}
 impl_int! {u32, 4}
+impl_int! {u64, 8}
 impl_int! {i32, 4}
+impl_int! {i64, 8}
```
claudiu725, line 54:
```suggestion
impl_int! {u8, 1}
impl_int! {u16, 2}
impl_int! {u32, 4}
impl_int! {u64, 8}
impl_int! {i8, 1}
impl_int! {i16, 2}
impl_int! {i32, 4}
impl_int! {i64, 8}
```
Should be symmetric in signed / unsigned -> added `i8`, `i16`

alyn509 (approved):

claudiu725 (approved):

elrond-wasm/src/types/managed/managed_vec_item.rs
andrei-marinica: Yes, true. Should also be tested. Remains TODO

PR #451, opened 2021-11-11
Implement `ExactSizeIterator` for `ManagedVecIterator`

claudiu725:
- add `size_hint` implementation, which is needed for `ExactSizeIterator`
- add `Clone` for `ManagedVecIterator`
- update macro replacement rules for `ManagedVecIterator` to allow skipping `Self::Api` when passing an iterator as a function argument
- add tests

### DIFF
--- a/elrond-wasm-debug/tests/managed_vec_test.rs
+++ b/elrond-wasm-debug/tests/managed_vec_test.rs
@@ -26,6 +26,25 @@ fn test_managed_vec_iter_rev() {
 }
 
 #[test]
+fn test_managed_vec_iter_exact_size_trait() {
+    let context = DebugApi::dummy();
+
+    let mut managed_vec = ManagedVec::new(context.clone());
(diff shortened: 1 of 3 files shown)


elrond-wasm-debug/tests/managed_vec_test.rs
```diff
@@ -25,6 +25,25 @@ fn test_managed_vec_iter_rev() {
...
 
+#[test]
+fn test_managed_vec_iter_exact_size_trait() {
+    let context = DebugApi::dummy();
+
+    let mut managed_vec = ManagedVec::new(context.clone());
+    for i in 1..=10 {
+        managed_vec.push(i);
+    }
+    assert!(managed_vec.len() == 10);
```
alyn509, line 36: some spacing between assert and variables would make this look better

alyn509 (approved):

PR #458, opened 2021-11-12
ManagedVarArgsEager

alyn509:


elrond-wasm/src/types/managed/managed_multi_result_vec_eager.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    pub fn append_vec(&mut self, item: ManagedMultiResultVecEager<M, T>) {
+        self.0.append_vec(item.0)
+    }
+
+    pub fn clear(&mut self) {
+        self.0.clear()
+    }
+
+    pub fn into_vec(self) -> Vec<T> {
```
dorin-iancu, line 76: Shouldn't this return a managed vec?

andrei-marinica: done

andrei-marinica (approved):

PR #460, opened 2021-11-12
Storage static operations

andrei-marinica:


andrei-marinica:
Changes have in the mean time become obsolete.

PR #471, opened 2021-11-19
Hex literal features

nu-wa:
- Adding [hex-literal](https://docs.rs/hex-literal/0.3.4/hex_literal/) as a feature to avoid having to include it since it is already a project dependency.
- Made elrond-wasm use [hex-literal](https://docs.rs/hex-literal/0.3.4/hex_literal/) version  `0.3.4` to avoid future possible breaking changes.

Inspired by [this example](https://github.com/ElrondNetwork/elrond-wasm-rs/blob/master/contracts/feature-tests/composability/proxy-test-first/src/proxy-test-first.rs#L8), my smart contract required a hardcoded address in it, thought this might be useful to others. Made it a feature so it's only optional.

### DIFF
--- a/elrond-wasm/src/lib.rs
+++ b/elrond-wasm/src/lib.rs
@@ -29,3 +29,6 @@ pub mod types;
 pub use hex_call_data::*;
 pub use io::*;
 pub use storage::{storage_clear, storage_get, storage_get_len, storage_set};
+
+#[cfg(feature="hex")]
+pub use hex_literal;
--- a/elrond-wasm/Cargo.toml
+++ b/elrond-wasm/Cargo.toml
@@ -23,6 +23,9 @@ hashbrown = "0.11.2"
 hex-literal = "0.3.1"
 bitflags = "1.3.2"
 
+[features]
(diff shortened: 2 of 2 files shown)


elrond-wasm/src/lib.rs
```diff
@@ -29,3 +29,6 @@ pub mod types;
 pub use hex_call_data::*;
 pub use io::*;
 pub use storage::{storage_clear, storage_get, storage_get_len, storage_set};
+
+#[cfg(feature="hex")]
```
andrei-marinica, line 33: Since we need the hex_literal crate in the framework, there is no need to guard it by a feature.
Other than that great idea, we will include it in the next release one way or another.
We also want to add some macro for hard-coded addresses, since it is sometimes useful.

andrei-marinica (changes requested):

andrei-marinica:
`hex` is now reexported.
The new release (elrond-wasm 0.25.0) has it.

nu-wa:
Thanks for the integration! Sorry I could do the requested changes I did not have time and kind of forgot about it

PR #472, opened 2021-11-22
Static API instance

andrei-marinica:


andrei-marinica:
Features implemented in `static-api-2`.

PR #478, opened 2021-11-23
managed address as managed buffer feature

andrei-diaconescu:


claudiu725:
Implemented in https://github.com/ElrondNetwork/elrond-wasm-rs/pull/477

PR #484, opened 2021-11-29
testing framework - generate mandos

dorin-iancu:
- generate mandos output for set blockchain/accounts state
- generate mandos output for SCCalls and SCQueries
- use relative paths in generated mandos


elrond-wasm-debug/src/testing_framework/contract_obj_wrapper.rs
```diff
@@ -35,16 +41,30 @@ where
     CB: ContractBase<Api = DebugApi> + CallableContract<DebugApi> + 'static,
     ContractObjBuilder: 'static + Copy + Fn(DebugApi) -> CB,
 {
-    pub fn new(_workspace_path: &str) -> Self {
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> Self {
+        let mut current_dir = std::env::current_dir().unwrap();
+        current_dir.push(PathBuf::from_str("mandos/").unwrap());
```
andrei-marinica, line 47: I think the final slash is not necessary, but doesn't hurt either

elrond-wasm-debug/src/testing_framework/mandos_generator.rs
```diff
@@ -0,0 +1,122 @@
...
+                name: None,
+                steps: Vec::new(),
+            },
+            current_tx_id: 0,
+        }
+    }
+
+    pub fn write_mandos_output(self, file_path: &str) {
+        let buf = Vec::new();
+        let formatter = serde_json::ser::PrettyFormatter::with_indent(b"    ");
```
andrei-marinica, line 31: We could extract this logic somewhere, I think it is also duplicated in some other tests.

elrond-wasm-debug/src/testing_framework/raw_converter.rs
```diff
@@ -0,0 +1,188 @@
...
+pub(crate) fn u64_as_raw(value: u64) -> ValueSubTree {
+    ValueSubTree::Str(value.to_string())
+}
+
+pub(crate) fn bytes_as_raw(bytes: &[u8]) -> ValueSubTree {
+    ValueSubTree::Str(bytes_to_hex(bytes))
+}
+
+pub(crate) fn bytes_to_hex(bytes: &[u8]) -> String {
+    "0x".to_owned() + &hex::encode(bytes)
```
andrei-marinica, line 187: I think something like `format!("0x{}", ...)` does this more elegantly, but this is also ok

claudiu725 (approved):

andrei-marinica (approved):

elrond-wasm-debug/src/testing_framework/raw_converter.rs
dorin-iancu: Good idea, always forget about format. Will fix in the next PR.

elrond-wasm-debug/src/testing_framework/mandos_generator.rs
dorin-iancu: I copied this from the serde test, so yeah, it's duplicated from there.

elrond-wasm-debug/src/testing_framework/contract_obj_wrapper.rs
dorin-iancu: iirc, PathBuff::push() doesn't work correctly without the slash, as it considers it an absolute path and overwrites the whole path with it.

PR #488, opened 2021-12-03
Only Owner module

alyn509:
all endpoints within this module have `#[only_owner]`


contracts/feature-tests/basic-features/mandos/only_owner.scen.json
```diff
@@ -61,6 +61,86 @@
                 "refund": "*"
             }
         },
+        {
```
andrei-marinica, line 64: Can we also move this test to use-module?
I see it was removed from the mandos-rs suite

dorin-iancu (approved):

contracts/feature-tests/use-module/src/only_owner_mod.rs
```diff
@@ -0,0 +1,9 @@
+elrond_wasm::imports!();
+
+/// This module is in the crate, but it is not included.
+/// Its endpoints should not appear in the contract binary.
+#[elrond_wasm::only_owner_module]
```
andrei-marinica, line 5: I would prefer to continue having
```
#[elrond_wasm::module]
#[only_owner]
```
Can we pull this off?

andrei-marinica (changes requested):

contracts/feature-tests/basic-features/mandos/only_owner.scen.json
alyn509: wrote the test initially in basic-features. 
now I placed it in its right place

elrond-wasm-derive/src/model/argument.rs
```diff
@@ -54,3 +54,14 @@ impl MethodArgument {
         matches!(self.metadata.payment, ArgPaymentMetadata::NotPayment)
     }
 }
+
+#[derive(Clone, Debug)]
+pub struct TraitProperties {
```
andrei-marinica, line 59: Clippy says this can be derived

elrond-wasm-derive/src/parse/method_parse.rs
```diff
@@ -30,7 +30,7 @@ pub fn process_method(m: &syn::TraitItemMethod) -> Method {
     let mut first_pass_data = MethodAttributesPass1 {
         method_name: m.sig.ident.to_string(),
         payable: MethodPayableMetadata::NotPayable,
-        only_owner: false,
+        only_owner: trait_attributes.only_owner.clone(),
```
andrei-marinica: Clippy: using `clone` on type `bool` which implements the `Copy` trait

andrei-marinica (approved):

PR #519, opened 2021-12-23
ManagedRef & ManagedReadonly, semantically correct

andrei-marinica:
Finally a semantically correct `ManagedRef`.
They act like proper references.
**They have `Copy` semantics.**
Copying them means copying the `i32` handle, which is cheap and makes sense. The handle is a pointer, semantically.

The old implementation was repurposed as `ManagedReadonly`.
It's like a smart pointer, with no deallocation.
`ManagedReadonly` objects are owned, but you cannot create a mutable instance from them without cloning (there is no `Rc::try_unwrap` functionality because they are not `Rc`s and there is no way to check the soundness of such an operation).
**They also have `Copy` semantics.**

The basic managed objects represent a mutable reference to the data (which lies in the VM). No additional mutable reference type created.

Under the new rules:
- `ManagedRef` obey Rust ownership rules via the lifetime, even though the lifetime is not used internally
- Creating a `ManagedReadonly` consumes the mutable object, so no mutable and immutable references to the same object can coexist.
- Managed type objects cannot be cloned without also cloning the underlying data, so new handle for each clone.


dorin-iancu (approved):
Some TODOs left, but otherwise, looks great

PR #527, opened 2021-12-30
Api refactor part 3 - StaticBufferApi refactored

andrei-marinica:


andrei-marinica:
Branch started from wrong point. Redoing branch & PR.

PR #541, opened 2022-01-05
Promises part 1

alyn509:


elrond-wasm/src/types/interaction/contract_call.rs
```diff
@@ -172,15 +198,21 @@ where
                 let zero = BigUint::zero();
                 let endpoint_name = ManagedBuffer::new_from_bytes(ESDT_TRANSFER_FUNC_NAME);
 
+                let success_callback = b"";
+                let error_callback = b"";
```
dorin-iancu, line 202: Why is this being reset? Same in other methods.

alyn509: good catch :+1: 
forgot so solve the esdt transfer calls

alyn509: this should remain only for `new_with_esdt_payment`

alyn509: just in case I will also make a mandos test with some transfers to make sure all works

andrei-marinica (approved):

PR #542, opened 2022-01-06
`signal_error!` macro with formatter

dorin-iancu:


elrond-wasm-derive/src/lib.rs
```diff
@@ -57,3 +63,84 @@ pub fn managed_vec_item_derive(input: proc_macro::TokenStream) -> proc_macro::To
...
+                let group_as_tt = arg_groups[arg_index].stream();
+                let mut group_tt_iter = group_as_tt.into_iter();
+                let ident = match group_tt_iter.next().unwrap() {
+                    proc_macro::TokenTree::Ident(ident) => ident,
+                    _ => panic!("Invalid argument, expected a named variable"),
+                };
+
+                let varname = format_ident!("{}", ident.to_string());
+                let encode_arg = quote! {
+                    #varname.top_encode(&mut ___encoded_arg___).unwrap();
```
andrei-marinica, line 118: Using fully qualified syntax here helps with hygiene.
I would suggest something like: `elrond_codec::TopEncode::top_encode(&#varname, &mut ___encoded_arg___)`

andrei-marinica (approved):

PR #552, opened 2022-01-13
managed vec get and get_mut

dorin-iancu:


alyn509 (dismissed):

andrei-marinica (approved):

elrond-wasm/src/types/managed/managed_vec_ref.rs
```diff
@@ -0,0 +1,66 @@
...
+use crate::api::{ErrorApiImpl, ManagedTypeApi};
+
+use super::{managed_vec::INDEX_OUT_OF_RANGE_MSG, ManagedVec, ManagedVecItem};
+
+pub struct ManagedVecRef<'a, M, T>
+where
+    M: ManagedTypeApi,
+    T: ManagedVecItem,
+{
+    managed_vec: &'a mut ManagedVec<M, T>,
```
andrei-marinica, line 12: Might be slightly more efficient wasm-wise to only store the handle here.
I will investigate.

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #562, opened 2022-01-23
Empty contract

andrei-marinica:
In order to have a truly empty contract template.


dorin-iancu (changes requested):
Since we're moving away from mandos to the Rust testing framework, I think it would be best to have a minimal setup of it instead of a mandos test.

Also add it to sizes report maybe?

dorin-iancu (approved):

PR #577, opened 2022-02-01
[TEST] Test `wasm-opt` using erdpy from the development branch

claudiu725:


claudiu725:
Close since the `wasm-opt` dependencies installation is present in `erdpy 1.0.25` which is now released. PR https://github.com/ElrondNetwork/elrond-wasm-rs/pull/582 will handle this now.

PR #594, opened 2022-02-14
Various improvements

dorin-iancu:
- new call_value `payment_as_tuple` method, that returns a tuple instead of constructing an `EsdtTokenPayment` struct
- improve `is_valid_esdt_identifier` implementation
- improve some SC examples to use the newest APIs and less dynamic allocation


elrond-codec/src/impl_for_types/impl_vec.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::transmute_undefined_repr)]
```
andrei-marinica, line 1: Only add this on the respective methods, not the entire crate, I think the warning can be useful.
I will investigate more if there are better solutions to the layout problem.

andrei-marinica (changes requested):

catalinnnn (approved):

claudiu725 (approved):

PR #628, opened 2022-03-05
Add possibility to set vote weight for board members

ctindogarus4f:
**Before this PR:**
All the board members have equal voting rights.

**After this PR:**
People might argue that a new employee should not have the same voting power as the CEO. To solve this issue, this PR introduces a new way of configuring the board, so that each board member can have different voting power.

**Nice to know:**
- by default, all members have a voting power of 1
- board members can add/edit the voting power of a member by proposing a `ChangeWeightForBoardMember` action
- in case of removal, the user and his voting power are not taken into consideration anymore
- the voting power doesn't have any cap, so the board is fully responsible for setting up the quorum and the weights properly for each member

**Regular flow:**
1. User creates a `multisig` SC, sets the quorum to 1 and only adds himself as a member in the board.
2. User adds all the other members in the board.
3. User sets voting power for each member.
4. User sets the quorum accordingly. (ex: if the board has 5 members and the voting power is 100, a recommended quorum would be 51)

**This change is backward compatible.**

**Left TODOs:**
- add `mandos` tests if this PR gets approval from Elrond community

### DIFF
--- a/contracts/examples/multisig/src/multisig.rs
+++ b/contracts/examples/multisig/src/multisig.rs
@@ -25,9 +25,8 @@ pub trait Multisig:
         let board_vec = board.to_vec();
         let new_num_board_members = self.add_multiple_board_members(board_vec);
 
-        let num_proposers = self.num_proposers().get();
         require!(
-            new_num_board_members + num_proposers > 0,
+            new_num_board_members > 0,
             "board cannot be empty on init, no-one would be able to propose"
         );
 
--- a/contracts/examples/multisig/src/multisig.rs
+++ b/contracts/examples/multisig/src/multisig.rs
@@ -158,7 +158,7 @@ pub trait Multisig:
             "only board members and proposers can discard actions"
         );
         require!(
-            self.get_action_valid_signer_count(action_id) == 0,
+            self.get_action_valid_signer_weight(action_id) == 0,
             "cannot discard action with valid signatures"
         );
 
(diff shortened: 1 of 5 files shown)


contracts/examples/multisig/src/multisig.rs
```diff
@@ -25,9 +25,8 @@ pub trait Multisig:
         let board_vec = board.to_vec();
         let new_num_board_members = self.add_multiple_board_members(board_vec);
 
-        let num_proposers = self.num_proposers().get();
         require!(
-            new_num_board_members + num_proposers > 0,
+            new_num_board_members > 0,
```
ctindogarus4f, line 29: side-by fix. `num_proposers` is always 0 on `init`

andrei-marinica: That condition is purposefully added to make the constructor upgrade-friendly. The `init` function is also called when upgrading, if anyone tries to upgrade it's likely they won't think about changing the constructor.

ctindogarus4f:
Let me guys know if you are happy with this implementation so I can do manual testing + add mandos tests + create a new PR holding this change, but in a separate directory, so it doesn't affect the current state of the multisig. (maybe `contracts/examples/multisig-weighted`?)

andrei-marinica:
We should create a separate contract with this feature. It can be made into a separate repo. The original multisig is considered stable as it is and audited, we will not be adding any new features to it.

andrei-marinica:
That being said, the implementation looks great, and we should definitely use it.

PR #629, opened 2022-03-06
Fix missing decorator on claim method in example sc

LegerG:
This decorator seems to be missing


andrei-marinica (approved):

andrei-marinica:
Please sign your commits in the future!

PR #639, opened 2022-03-14
Multisig rust tests

dorin-iancu:
- Tests for all actions except deploy & upgrade


contracts/examples/multisig/tests/multisig_rust_test_setup/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+            &self.ms_wrapper,
+            &rust_biguint!(0),
+            |sc| {
+                let _ = sc.discard_action(action_id);
+            },
+        )
+    }
+}
+
+fn boxed_bytes_vec_to_managed<M: ManagedTypeApi>(
```
andrei-marinica, line 271: I think we have conversion already implemented for that, try .into() or ManagedVec::<M, ManagedBuffer<M>>::from(raw_vec)

andrei-marinica: You can find the From implementation in managed_vec.rs line 78

andrei-marinica (changes requested):

andrei-marinica (dismissed):

contracts/examples/multisig/tests/multisig_rust_test_setup/mod.rs
dorin-iancu: The value required is MultiValueEncoded, which is why I convert separately, and then use `into()` again on the result

catalinnnn (approved):

PR #659, opened 2022-03-31
Bi di mapper

alyn509:


elrond-wasm/src/storage/mappers/bi_di_mapper.rs
```diff
@@ -0,0 +1,269 @@
...
+        self.id_set_mapper.insert(id);
+        self.value_set_mapper.insert(value);
+        true
+    }
+
+    pub fn remove_by_id(&mut self, id: &K) -> bool {
+        if self.id_set_mapper.swap_remove(id) {
+            let value = self.get_value(id);
+            storage_clear(self.get_id_key(&value).as_ref());
+            storage_clear(self.get_value_key(id).as_ref());
```
dorin-iancu, line 151: Extract those into private methods, something like `clear_id()` and `clear_value()` respectively.

elrond-wasm/src/storage/mappers/bi_di_mapper.rs
```diff
@@ -0,0 +1,269 @@
...
+            self.value_set_mapper.swap_remove(&value);
+            return true;
+        }
+        false
+    }
+    pub fn remove_by_value(&mut self, value: &V) -> bool {
+        if self.value_set_mapper.swap_remove(value) {
+            let id = self.get_id(value);
+            storage_clear(self.get_id_key(value).as_ref());
+            storage_clear(self.get_value_key(&id).as_ref());
```
dorin-iancu, line 161: Same comment as above.

elrond-wasm/src/storage/mappers/bi_di_mapper.rs
```diff
@@ -0,0 +1,269 @@
...
+        result
+    }
+
+    pub fn get_all_ids(&self) -> ManagedVec<SA, K> {
+        let mut result = ManagedVec::new();
+        for id in self.id_set_mapper.iter() {
+            result.push(id);
+        }
+        result
+    }
```
dorin-iancu, line 200: Maybe these should return an iterator of some kind instead of manually pushing them into a ManagedVec.

elrond-wasm/src/storage/mappers/bi_di_mapper.rs
```diff
@@ -0,0 +1,269 @@
...
+        + NestedEncode
+        + NestedDecode
+        + 'static
+        + Default
+        + PartialEq
+        + ManagedVecItem,
+{
+    type DecodeAs = MultiValueEncoded<SA, V>;
+
+    fn multi_encode_or_handle_err<O, H>(&self, output: &mut O, h: H) -> Result<(), H::HandledErr>
```
dorin-iancu, line 229: I think this should return all (key, value) pairs instead of only the values.

elrond-wasm/src/storage/mappers/bi_di_mapper.rs
```diff
@@ -0,0 +1,266 @@
...
+where
+    SA: StorageMapperApi,
+    K: TopEncode
+        + TopDecode
+        + NestedEncode
+        + NestedDecode
+        + 'static
+        + Default
+        + PartialEq
+        + ManagedVecItem,
```
andrei-marinica, line 95: ManagedVecItem is not necessary, we should yield an iterator, not push it into a ManagedVec.
If we provide a method to extract all keys/values directly to a ManagedVec, the trait restriction should be only around that method.

elrond-wasm/src/storage/mappers/bi_di_mapper.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    pub fn get_all_ids(&self) -> ManagedVec<SA, K> {
+        let mut result = ManagedVec::new();
+        for id in self.id_set_mapper.iter() {
+            result.push(id);
+        }
+        result
+    }
+
+    pub fn len(&self) -> usize {
```
andrei-marinica, line 202: clippy :)

elrond-wasm/src/storage/mappers/bi_di_mapper.rs
andrei-marinica: Yes, same as for the MapMapper

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #664, opened 2022-04-04
Constructed Mandos tests using Proxies (prototype)

andrei-marinica:


elrond-wasm-debug/src/mandos_step/quick_call.rs
```diff
@@ -0,0 +1,72 @@
...
+            args: contract_call.arg_buffer.to_raw_args_vec(),
+            gas_limit: u64::MAX,
+            gas_price: 0u64,
+            tx_hash: H256::zero(),
+        };
+
+        let tx_result = self.with_borrowed_rc(|rc| sc_query(tx_input, rc.clone()));
+        assert!(tx_result.result_status == 0, "quick query failed"); // TODO: print more
+        assert!(
+            tx_result.result_status != 0 || tx_result.result_calls.is_empty(),
```
dorin-iancu, line 37: This condition seems a bit weird. Shouldn't we only assert result calls? Since we should assert status == 0 for success.
```suggestion
            tx_result.result_calls.is_empty(),
```

dorin-iancu (approved):

PR #673, opened 2022-04-09
ESDT transfer with fee contract

alyn509:


contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        let caller = self.blockchain().get_caller();
+        self.send().direct_multi(&caller, &fees, &[]);
+    }
+
+    #[payable("*")]
+    #[endpoint]
+    fn transfer(&self, address: ManagedAddress) {
+        require!(!self.general_fee().is_empty(), "Fees are not set yet");
+        let payments = self.call_value().all_esdt_transfers();
```
andrei-marinica, line 40: What happens when someone is sending EGLD? Should the fees apply or should it be rejected?

andrei-marinica (changes requested):

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
alyn509: tokens that have no fee will be simply transfered withour expecting a fee

contracts/examples/esdt-transfer-with-fee/src/fee.rs
```diff
@@ -0,0 +1,13 @@
+elrond_wasm::imports!();
+elrond_wasm::derive_imports!();
+
+pub(crate) const PERCENTAGE_DIVISOR: u32 = 10000; // dividing the percentage fee by this number will result in a 4 decimal percentage
```
dorin-iancu, line 4:
```suggestion
pub(crate) const PERCENTAGE_DIVISOR: u32 = 10_000; // dividing the percentage fee by this number will result in a 4 decimal percentage
```

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,105 @@
...
+
+elrond_wasm::imports!();
+#[elrond_wasm::contract]
+pub trait EsdtTransferWithFee {
+    #[init]
+    fn init(&self) {}
+
+    #[only_owner]
+    #[endpoint(setFeeForToken)]
+    fn set_fee_for_token(&self, fee: Fee<Self::Api>, token: TokenIdentifier) {
```
dorin-iancu, line 14: I think this should be split into two endpoints, one for percentage, one for fixed. It's pretty difficult to encode the arguments otherwise.

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,105 @@
...
+        self.token_fee(&token).set(fee);
+    }
+
+    #[only_owner]
+    #[endpoint(claimFees)]
+    fn claim_fees(&self) {
```
dorin-iancu, line 20: Should check if `self.paid_fees()` is empty or `fees` vec is empty.

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,105 @@
...
+            fees.push(EsdtTokenPayment::new(token, nonce, amount))
+        }
+        self.paid_fees().clear();
+
+        let caller = self.blockchain().get_caller();
+        self.send().direct_multi(&caller, &fees, &[]);
+    }
+
+    #[payable("*")]
+    #[endpoint(transfer)]
```
dorin-iancu, line 32: Unneeded name in annotation. `#[endpoint]` is enough when the name is identical

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
dorin-iancu: This will not work for EGLD, as it's not transferable through the multi-transfer API.

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,105 @@
...
+                        payment.amount >= fee.amount,
+                        "Insufficient payments for covering fees"
+                    );
+
+                    self.paid_fees()
+                        .entry((payment.token_identifier.clone(), payment.token_nonce))
+                        .or_insert(0u64.into())
+                        .update(|value| *value += fee.amount.clone());
+
+                    payment.amount -= fee.amount.clone();
```
dorin-iancu, line 58:
```suggestion
                    payment.amount -= &fee.amount;
```

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,105 @@
...
+                if payment.token_identifier == fee.token_identifier {
+                    require!(
+                        payment.amount >= fee.amount,
+                        "Insufficient payments for covering fees"
+                    );
+
+                    self.paid_fees()
+                        .entry((payment.token_identifier.clone(), payment.token_nonce))
+                        .or_insert(0u64.into())
+                        .update(|value| *value += fee.amount.clone());
```
dorin-iancu, line 56:
```suggestion
                        .update(|value| *value += &fee.amount);
```

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,105 @@
...
+                    payment.amount -= fee.amount.clone();
+                    perceived_tax = true;
+                    break;
+                }
+            }
+
+            require!(perceived_tax, "Fee payment missing");
+        }
+
+        self.send().direct_multi(&address, &payments, &[]);
```
dorin-iancu, line 67: This will fail if `payment.amount == fee.amount` in the previous `for`, as you'd have amounts of `0`. I think you should do a post-processing here on the Vec and create a new one with only the non-zero values. And only send if the new vec is not empty.

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,105 @@
...
+            Fee::ExactValue(requested) => calculated_fee = requested,
+
+            Fee::Percentage(percentage) => {
+                let calculated_fee_amount =
+                    provided.amount.clone() * percentage / PERCENTAGE_DIVISOR;
+                provided.amount = calculated_fee_amount;
+                calculated_fee = provided;
+            },
+        }
+        calculated_fee
```
dorin-iancu, line 95:
```suggestion
        match fee {
            Fee::ExactValue(requested) => requested,
            Fee::Percentage(percentage) => {
                let calculated_fee_amount =
                    &provided.amount * &percentage / PERCENTAGE_DIVISOR;
                provided.amount = calculated_fee_amount;
                
                provided
            },
        }
```

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,105 @@
...
+        self.send().direct_multi(&address, &payments, &[]);
+    }
+
+    fn get_fee(&self, mut payment: EsdtTokenPayment<Self::Api>) -> EsdtTokenPayment<Self::Api> {
+        if self.token_fee(&payment.token_identifier).is_empty() {
+            payment.amount = 0u64.into();
+            payment
+        } else {
+            self.calculate_fee(self.token_fee(&payment.token_identifier).get(), payment)
+        }
```
dorin-iancu, line 76: This builds the mapper twice. It can be rewritten as:
```suggestion
        let fee_mapper = self.token_fee(&payment.token_identifier);
        if fee_mapper.is_empty() {
            payment.amount = 0u64.into();
            payment
        } else {
            self.calculate_fee(fee_mapper.get(), payment)
        }
```

dorin-iancu (changes requested):

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
alyn509: ^ no implementation for `&elrond_wasm::types::BigUint<<Self as ContractBase>::Api> * &u32`

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
alyn509: ^ no implementation for `&elrond_wasm::types::BigUint<<Self as ContractBase>::Api> * &u32`

alyn509: :dogwithtears

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,133 @@
...
+        new_payment
+    }
+
+    fn get_fee_type(&self, payment: &EsdtTokenPayment<Self::Api>) -> Fee<Self::Api> {
+        let fee_mapper = self.token_fee(&payment.token_identifier);
+        if fee_mapper.is_empty() {
+            Fee::Unset
+        } else {
+            fee_mapper.get()
+        }
```
dorin-iancu, line 104: Since empty storage will yield "Unset", you can safely `.get` without checking for empty. So this whole function can be removed and you can use `self.token_fee(&payment.token_identifier).get()` directly.

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+        self.send().direct_multi(&address, &new_payments, &[]);
+    }
+
+    fn get_payment_after_fees(
+        &self,
+        fee: Fee<Self::Api>,
+        payment: &EsdtTokenPayment<Self::Api>,
+    ) -> EsdtTokenPayment<Self::Api> {
```
dorin-iancu, line 82: It doesn't seem like you use the result anywhere, so you could probably just remove the returned payment.

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,133 @@
...
+                },
+                Fee::Unset => {
+                    new_payments.push(payment);
+                },
+            }
+        }
+        self.send().direct_multi(&address, &new_payments, &[]);
+    }
+
+    fn get_payment_after_fees(
```
dorin-iancu, line 78: If you follow the next comment, I suggest renaming to something else, as you're not really "getting" anything anymore.

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,133 @@
...
+                        next_payment.amount == fee.amount,
+                        "Insufficient payments for covering fees"
+                    );
+
+                    self.get_payment_after_fees(fee_type, &next_payment);
+                    new_payments.push(payment);
+                },
+                Fee::Percentage(_) => {
+                    self.get_payment_after_fees(fee_type, &payment);
+                    new_payments.push(payment);
```
dorin-iancu, line 68: Hm, it seems like you don't cut anything from the original payment here.

dorin-iancu: So you should probably either use the payment returned from `get_payment_after_fees`, or give a mutable reference to the function so it modifies that directly.

dorin-iancu (approved):

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
```diff
@@ -0,0 +1,122 @@
...
+
+elrond_wasm::imports!();
+#[elrond_wasm::contract]
+pub trait EsdtTransferWithFee {
+    #[init]
+    fn init(&self) {}
+
+    #[only_owner]
+    #[endpoint(setExactValueFee)]
+    fn set_exact_value_fee(&self, fee: EsdtTokenPayment<Self::Api>, token: TokenIdentifier) {
```
andrei-marinica, line 14: Better split the fee argument into token/value (they're all fungible, right? so no nonce), to make it easier to call the endpoint

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
andrei-marinica: EGLD condition still missing. Try something like
`require!(self.call_value().egld_value() == 0, "EGLD transfers not allowed);`

Better framework syntax coming soon, but until then ....

andrei-marinica (changes requested):

andrei-marinica (approved):

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
sasurobert: +1

contracts/examples/esdt-transfer-with-fee/src/esdt_transfer_with_fee.rs
sasurobert: +1

PR #677, opened 2022-04-11
Multisig tests

claudiu725:
- Add tests for the multisig smart contract which use the rust testing framework v2
- Add several `CodecFrom` implementations
- Introduce `CodecSelf` for allowing other types to override the default `CodecFrom` implementation
- Add `CodecFrom` implementations for `OptionalValue` to allow:
  - to use an `OptionalValue` whose inner type is compatible with the original type (implements `CodecFrom`)
  - to use a `value` directly instead of an `OptionalValue::Some(value)`
- Add `ManagedArgBuffer::into_multi_value_encoded` to allow using the proxy-generated argument buffer as a multi-value in order to forward the arguments to multisig


dorin-iancu (dismissed):

contracts/examples/multisig/tests/multisig_rust_test_v2.rs
```diff
@@ -0,0 +1,319 @@
...
+
+#[test]
+fn multisig_adder_test() {
+    let _ = DebugApi::dummy();
+
+    let world = &mut world();
+    let ic = &world.interpreter_context();
+
+    let users = setup_users(world, ic);
+    let source_adder = &mut adder_deploy(world, ic);
```
andrei-marinica, line 71: `ic` is part of `world`, so I don't think we need it as a separate argument

contracts/examples/multisig/tests/multisig_rust_test_v2.rs
```diff
@@ -0,0 +1,319 @@
...
+        world,
+        ic,
+    );
+    let expected_adder_address = AddressKey::interpret_from("sc:adder-multisig", ic);
+    assert_eq!(
+        adder.mandos_address_expr.value,
+        expected_adder_address.value
+    );
+
+    let first_value: BigUint = 42u64.into();
```
andrei-marinica, line 89: Couldn't we actually only work with u64 or even u32 up to the contract proxy? There will unlikely be any big numbers in the test.

contracts/examples/multisig/tests/multisig_rust_test_v2.rs
```diff
@@ -0,0 +1,319 @@
...
+) -> Option<Address> {
+    multisig_sign_multiple(action_id, signers, multisig, world);
+    multisig_perform(action_id, caller, multisig, world)
+}
+
+fn multisig_deploy_adder(
+    multisig: &mut MultisigContract,
+    source_adder: &mut AdderContract,
+    caller: &Address,
+    signers: &[&Address],
```
andrei-marinica, line 227: Would probably be nice to group these arguments in a context object.

elrond-codec/src/equivalent/codec_from.rs
```diff
@@ -7,7 +7,9 @@ where
 {
 }
 
-impl<T> CodecFrom<T> for T where T: TopEncodeMulti + TopDecodeMulti {}
+pub auto trait CodecSelf {}
```
andrei-marinica, line 10: Would rename to `CodecFromSelf`.

andrei-marinica (changes requested):

andrei-marinica (approved):

andrei-marinica (approved):

PR #687, opened 2022-04-14
Formatter: Add Codec and Bytes suport

CostinCarabas:
Allows to print("{:c}", var) and print("{":b}", var)


CostinCarabas:
@andrei-marinica , As discussed in the last commit, I:
* Added tests for ascii, binary, hex and codecs.
* Refactored *_bytes -> *_binary
* Removed overflow issue for binary format

CostinCarabas:
now with a Verified commit :sunglasses: 
image

andrei-marinica (approved):

PR #692, opened 2022-04-21
cleanup return data

dorin-iancu:
- cleanup return data VM function integration
- execute on dest context now automatically cleans up propagated results from the final output
- fixed tests


contracts/feature-tests/composability/forwarder/src/call_sync.rs
```diff
@@ -7,74 +7,6 @@ pub trait ForwarderSyncCallModule {
     #[proxy]
     fn vault_proxy(&self) -> vault::Proxy<Self::Api>;
 
-    #[endpoint]
-    #[payable("*")]
-    fn echo_arguments_sync(
```
andrei-marinica, line 12: Please don't delete this, it is still a valid test.

contracts/feature-tests/composability/mandos/forw_raw_sync_echo.scen.json
```diff
@@ -1,121 +0,0 @@
...
-                "sc:forwarder": {
-                    "nonce": "0",
-                    "balance": "0",
-                    "code": "file:../forwarder-raw/output/forwarder-raw.wasm"
-                }
-            }
-        },
-        {
-            "step": "scCall",
-            "txId": "call_execute_on_dest_context",
```
andrei-marinica, line 25: Please don't delete this file, it is still a valid test.

andrei-marinica (changes requested):

contracts/feature-tests/composability/forwarder/src/call_sync.rs
dorin-iancu: restored

contracts/feature-tests/composability/mandos/forw_raw_sync_echo.scen.json
dorin-iancu: restored

elrond-wasm-debug/src/api/send_api_mock.rs
```diff
@@ -577,6 +549,22 @@ impl SendApiImpl for DebugApi {
             arg_buffer.to_raw_args_vec(),
         );
 
+        self.clean_return_data();
+
         ManagedVec::from(result)
     }
+
+    fn clean_return_data(&self) {
```
andrei-marinica, line 557: (optional) `clear_return_data` I think would be a better name

elrond-wasm-node/src/api/send_api_node_impl_managed.rs
```diff
@@ -398,49 +406,9 @@ impl SendApiImpl for VmApiImpl {
                 result_handle,
             );
 
-            ManagedVec::from_raw_handle(result_handle)
-        }
-    }
+            self.clean_return_data();
```
andrei-marinica, line 409: I would like us to keep the API traits as close to the actual VM as possible.
That is why I would prefer to have this called higher-level, in the ContractCall, ContractDeploy and SendWrapper.
Probably the best way to implement would be to have an `execute_on_dest_context_raw` method in the SendWrapper and have ContractCall, ContractDeploy call it (this is some technical debt we need to sort out - the send api is doing too much).

On a related note: forwarder-raw is calling the API directly. This is also a problem, we should discourage this, it should also go though `self.send().execute_on_dest_context_raw(...)`.

Another technical debt: `call_local_esdt_built_in_function` does not belong in the API.

andrei-marinica (approved):

andrei-marinica (approved):

PR #708, opened 2022-05-10
fix typeabi impl for bidimapper

alyn509:


dorin-iancu (changes requested):
Remove the expanded version of the egld-esdt-swap

dorin-iancu (approved):

PR #710, opened 2022-05-10
Async to Async payment in Callback

dorin-iancu:


andrei-marinica:
Issue fixed in https://github.com/ElrondNetwork/elrond-wasm-rs/pull/892

PR #711, opened 2022-05-12
events in abi

alyn509:


elrond-wasm-debug/src/abi_json/event_abi_json.rs
```diff
@@ -0,0 +1,59 @@
...
+        EventInputAbiJson {
+            arg_name: abi.arg_name.to_string(),
+            type_name: abi.type_name.clone(),
+            indexed: if abi.indexed { Some(true) } else { None },
+        }
+    }
+}
+
+/// Same as EventAbiJson but ignores the name
+#[derive(Serialize, Deserialize)]
```
andrei-marinica, line 27: What do we need this for? There is no separate type of event for constructors.

contracts/feature-tests/abi-tester/abi_tester_expected_main.abi.json
```diff
@@ -391,6 +391,7 @@
             "outputs": []
         }
     ],
+    "events": [],
```
andrei-marinica, line 394: Add events in the abi tester contract to check that they are rendered correctly.

elrond-wasm-derive/src/model/endpoint.rs
```diff
@@ -33,5 +38,6 @@ pub enum PublicRole {
     CallbackRaw,
 
     /// Can only called from within the smart contract.
+    Event(EventMetadata),
```
andrei-marinica, line 41: Please don't change the model like this. Events are not public, since they cannot be called from the outside.
The ABI generator can figure out that it is an event from the fact that is has an event auto-implementation.
Sorry, but this part will need to be reverted.

elrond-wasm-derive/src/generate/abi_gen.rs
```diff
@@ -100,6 +100,52 @@ fn generate_endpoint_snippets(contract: &ContractTrait) -> Vec<proc_macro2::Toke
...
+        #(#input_snippets)*
+    }
+}
+
+fn generate_event_snippets(contract: &ContractTrait) -> Vec<proc_macro2::TokenStream> {
+    contract
+        .methods
+        .iter()
+        .filter_map(|m| match &m.public_role {
+            PublicRole::Event(event_metadata) => {
```
andrei-marinica, line 136: Replace with something like:
`if let MethodImpl::Generated(AutoImpl::Event{identifier}) = m.implementation { ... `

Event is not a PublicRole, since it is not publicly callable.
... maybe we should rename PublicRole to PubliclyCallable to avoid future confusions

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #714, opened 2022-05-13
pretty address format for generated mandos

dorin-iancu:
Generated mandos files now use the proper "address:" and "sc:" formats for addresses instead of hex. Hex format will still be used in cases where the address is not interpretable as a string.


elrond-wasm-debug/src/world_mock/blockchain_mock.rs
```diff
@@ -18,6 +13,7 @@ const ELROND_REWARD_KEY: &[u8] = b"ELRONDreward";
 #[derive(Debug)]
 pub struct BlockchainMock {
     pub accounts: HashMap<Address, AccountData>,
+    pub addr_to_mandos_string_map: HashMap<Address, String>,
```
andrei-marinica, line 16: We could wrap this into an object, for future extensions, but for now it looks great

andrei-marinica (approved):

PR #719, opened 2022-05-19
unique id mapper

dorin-iancu:
A unique ID mapper that holds numbers from 1 to N. If index is equal to entry value, then "nothing" is stored, so initializing the mapper is an O(1) operation.


elrond-wasm/src/storage/mappers/unique_id_mapper.rs
```diff
@@ -0,0 +1,189 @@
...
+        if id == EMPTY_ENTRY {
+            index
+        } else {
+            id
+        }
+    }
+
+    /// Gets the value from the index and removes it.
+    /// The value is replaced by the last item, and length is decremented.
+    pub fn get_and_swap_remove(&mut self, index: usize) -> UniqueId {
```
alyn509, line 76: maybe rename it `pop`

alyn509 (approved):

PR #721, opened 2022-05-19
sparse array

dorin-iancu:
- an array in which has the property that if array[i] == i, then the value 0 is stored


contracts/feature-tests/basic-features/src/sparse_array_features.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    #[endpoint]
+    fn sparse_array_set(
+        &self,
+        array: SparseArray<Self::Api, 100>,
+        index: usize,
+        value: usize,
+    ) -> SparseArray<Self::Api, 100> {
+        let mut array_clone = array;
+        array_clone.set(index, value);
```
andrei-marinica, line 35: would rather name this `mut_array`, it is not a clone

dorin-iancu: It was originally actually `array_clone = array.clone()`, but Clippy complained. Renamed now.

andrei-marinica (approved):

PR #729, opened 2022-05-23
checked single ESDT transfer

dorin-iancu:


elrond-wasm/src/contract_base/wrappers/call_value_wrapper.rs
```diff
@@ -58,10 +58,12 @@ where
     /// Warning, not tested with multi transfer, use `all_esdt_transfers` instead!
```
andrei-marinica, line 58: You made this content deprecated, so can be changed.
Point out that this crashes for multi-transfer.

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #749, opened 2022-06-08
Fix async cross shard result check

dorin-iancu:


contracts/feature-tests/composability/forwarder/src/call_async.rs
```diff
@@ -17,6 +17,42 @@ pub trait ForwarderAsyncCallModule {
     #[proxy]
     fn vault_proxy(&self) -> vault::Proxy<Self::Api>;
 
+    #[endpoint]
+    fn echo_args_async(&self, to: ManagedAddress, args: MultiValueEncoded<ManagedBuffer>) {
```
andrei-marinica, line 21: rename to forward_echo_arguments so it is clearer what we are doing

andrei-marinica (approved):

contracts/feature-tests/composability/forwarder/src/call_async.rs
dorin-iancu: For not I'll leave it like this, since it's already being used by some tests in VM as far as I know.

PR #761, opened 2022-06-23
only user account annotation

dorin-iancu:
#[only_user_account] makes it so the endpoint cannot be called by other smart contracts


claudiu725 (approved):

elrond-wasm/src/contract_base/wrappers/blockchain_wrapper.rs
```diff
@@ -81,6 +81,12 @@ where
         }
     }
 
+    pub fn check_caller_is_user_account(&self) {
```
andrei-marinica, line 84: can be done more efficiently by going directly to the API:
```
A::blockchain_api_impl().load_caller_managed(const_handles::MB_TEMP_1);
A::blockchain_api_impl().is_smart_contract(const_handles::MB_TEMP_1)
```
Avoids creating a new handle

andrei-marinica (changes requested):

elrond-wasm/src/contract_base/wrappers/blockchain_wrapper.rs
```diff
@@ -81,6 +81,12 @@ where
...
+        if self.is_smart_contract(&self.get_caller()) {
+            A::error_api_impl().signal_error(b"Endpoint can only be called by user accounts");
```
andrei-marinica, line 86: Also please put this in err_msg.rs

PR #769, opened 2022-06-27
mandos typed calls

andrei-marinica:
Bonus: some nicer syntax when writing mandos-v2.


elrond-wasm-debug/src/mandos_system/model/step/step_sc_call.rs
```diff
@@ -87,4 +95,78 @@ impl ScCallStep {
...
+    }
+}
+
+/// Extracts
+/// - recipient,
+/// - endpoint name,
+/// - the arguments.
+pub(super) fn process_contract_call<OriginalResult>(
+    contract_call: ContractCall<DebugApi, OriginalResult>,
+) -> (String, String, Vec<String>) {
```
dorin-iancu, line 144: I tend to avoid tuples when some of the fields have the same type, as it's easy to make a mistake and return/use them in the wrong order. I'd suggest a wrapper struct, something like "ContractCallRawString".

alyn509 (approved):

dorin-iancu (approved):

elrond-wasm-debug/src/mandos_system/model/step/step_sc_call.rs
andrei-marinica: you're right, but it's private and will stay so, it is just used internally for code deduplication

PR #775, opened 2022-07-04
Nft merge refactor

dorin-iancu:
- Add a limit of (currently) 10 NFTs merged into a single one
- Keep the list of merged tokens sorted by token ID, and then nonce
- Consolidate same token ID and nonce instances under a single entry 
- Use ArrayVec instead of ManagedVec and TokenIdentifier wherever possible - this allows us to sort, search, insert, and merge positions more easily instead of comparing managed types, which require API calls


elrond-wasm-modules/src/token_merge/mod.rs
```diff
@@ -73,40 +78,51 @@ pub trait TokenMergeModule:
...
 
+        let mut already_merged_tokens_iter = already_merged_tokens.into_iter();
+        let mut merged_attributes = if let Some(already_merged) = already_merged_tokens_iter.next()
+        {
+            MergedTokenAttributes::new_from_sorted_instances(already_merged)
+        } else {
+            MergedTokenAttributes::new()
+        };
+
+        while let Some(already_merged) = already_merged_tokens_iter.next() {
```
andrei-marinica, line 107: clippy

andrei-marinica (approved):

psorinionut (approved):

PR #781, opened 2022-07-07
storage raw api

dorin-iancu:
- storage read raw by key
- storage write raw by key
- storage read from another address


elrond-wasm/src/contract_base/wrappers/storage_raw_wrapper.rs
```diff
@@ -0,0 +1,74 @@
...
+        StorageRawWrapper {
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Reads storage from the given key, and deserializes the value to the provided type.
+    ///
+    /// Use only if really necessary, storage mappers should be preferred.
+    #[inline]
+    pub fn read<T: TopDecode>(&self, storage_key: &StorageKey<A>) -> T {
```
andrei-marinica, line 39: Simple to implement and it helps with syntax: having the key of generic type K, where StorageKey<A>: From<K>, and calling .into() inside.

Also implement `StorageKey` `From` `&str`.

We can then write `self.storage_raw().read("my_key")` and `self.storage_raw().read(managed_buffer)`

PR #788, opened 2022-07-12
Add overflow checks for integers

claudiu725:
- created `set_value` methods in both `BigInt` and `BigUint` which replace direct calls of `bi_set_int64` and provide overflow checks, which may occur when type casting from an `u64` that would normally become negative when interpreted as `i64`
- changed type of the exponent of the `pow` method of `BigFloat` from `u32` to `i32`
- added mandos tests which go over the overflow edge cases
- note: in the `from` implementation of `BigFloat`, the `as i64` cast is safe for the currently implemented types


elrond-wasm/src/types/managed/basic/big_int.rs
```diff
@@ -121,7 +145,7 @@ impl<M: ManagedTypeApi> BigInt<M> {
     pub fn zero() -> Self {
         let handle: M::BigIntHandle = M::static_var_api_impl().next_handle();
         // TODO: seting 0 will no longer be needed once we fix VM handle error
-        M::managed_type_impl().bi_set_int64(handle.clone(), 0);
+        Self::set_value(handle.clone(), 0);
```
dorin-iancu, line 148: Since setting to 0 is always safe, I think we can just keep calling the low-level API. No need to do all the overflow checks for 0.

elrond-wasm/src/types/managed/basic/big_uint.rs
```diff
@@ -110,7 +136,7 @@ impl<M: ManagedTypeApi> BigUint<M> {
     #[inline]
     pub fn zero() -> Self {
         let handle: M::BigIntHandle = M::static_var_api_impl().next_handle();
-        M::managed_type_impl().bi_set_int64(handle.clone(), 0);
+        Self::set_value(handle.clone(), 0u32);
```
dorin-iancu, line 139: Same as for BigInt, no need to go through all the checks

elrond-wasm/src/types/managed/basic/big_int_cmp.rs
```diff
@@ -29,31 +29,33 @@ impl<M: ManagedTypeApi> Ord for BigInt<M> {
...
 }
 
-fn cmp_i64<M: ManagedTypeApi>(bi: &BigInt<M>, other: i64) -> Ordering {
+fn cmp<M: ManagedTypeApi, T>(bi: &BigInt<M>, other: T) -> Ordering
+where
+    T: TryInto<i64> + PartialEq<T> + From<u8>,
+{
     let api = M::managed_type_impl();
-    if other == 0 {
+    if other == 0u8.into() {
```
andrei-marinica, line 37: Isn't it more natural to convert to i64 the first thing and then use the i64 from there on?

I would even split this function in 2:
fn cmp_i64(..., other: i64)
fn cmp_conv_i64<T: TryInto<i64>>(..., other: T) 

I know there's a bit of refactoring there, but it's not that complicated

andrei-marinica: Actually I think we should be unifying this logic with BigUint, it's pretty much the same code

elrond-wasm/src/types/managed/basic/big_int.rs
claudiu725: Done.

elrond-wasm/src/types/managed/basic/big_uint.rs
claudiu725: Done.

elrond-wasm/src/types/managed/basic/big_uint_cmp.rs
```diff
@@ -29,17 +29,19 @@ impl<M: ManagedTypeApi> Ord for BigUint<M> {
...
 }
 
-fn cmp_i64<M: ManagedTypeApi>(bi: &BigUint<M>, other: i64) -> Ordering {
+fn cmp<M: ManagedTypeApi, T>(bi: &BigUint<M>, other: T) -> Ordering
+where
+    T: TryInto<i64> + PartialEq<T> + From<u8>,
+{
     let api = M::managed_type_impl();
-    if other == 0 {
+    if other == 0u8.into() {
```
andrei-marinica, line 37: Isn't it more natural to convert to i64 the first thing and then use the i64 from there on?

I would even split this function in 2:
fn cmp_i64(..., other: i64)
fn cmp_conv_i64<T: TryInto>(..., other: T)

I know there's a bit of refactoring there, but it's not that complicated


Also - unite with the one from big int.

andrei-marinica (changes requested):

elrond-wasm/src/types/managed/basic/big_int_cmp.rs
claudiu725: Done. Implemented as suggested.

elrond-wasm/src/types/managed/basic/big_uint_cmp.rs
claudiu725: Merged with the other implementation.

elrond-wasm-debug/src/api/managed_types/big_int_api_mock.rs
```diff
@@ -185,7 +186,7 @@ impl BigIntApi for DebugApi {
     fn bi_pow(&self, dest: Self::BigIntHandle, x: Self::BigIntHandle, y: Self::BigIntHandle) {
         let bi_x = self.bi_get(x);
         let bi_y = self.bi_get(y);
-        let exp = big_int_to_i64(&bi_y).unwrap() as usize;
+        let exp = big_int_to_i64(&bi_y).unwrap().try_into().unwrap();
```
andrei-marinica, line 189: Unwrap `panic`s. We need to `signal_error`.

`let exp = big_int_to_i64(&bi_y).unwrap().try_into().unwrap_or_else(...)`

elrond-wasm/src/types/managed/basic/big_num_cmp.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use super::{cast_to_i64::cast_to_i64, BigInt};
+
+pub(crate) fn cmp_i64<M, B>(bi: &B, other: i64) -> Ordering
+where
+    M: ManagedTypeApi,
+    B: ManagedType<M, OwnHandle = M::BigIntHandle>,
+{
+    let api = M::managed_type_impl();
+    if other == 0u8.into() {
```
andrei-marinica, line 16: `other == 0i64`

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #799, opened 2022-07-25
added missing require

rvignolo:
It seems that the `require_setup_period_live();` method should be called at `end_setup_period`. Otherwise, this method could be called by the owner multiple times and more tokens will be minted.


psorinionut:
Great find, thanks for your contribution!

PR #800, opened 2022-07-29
feat(actionHistory): add log event on the perform action transaction

raresraulea:


contracts/examples/multisig/src/multisig_perform.rs
```diff
@@ -87,9 +87,24 @@ pub trait MultisigPerformModule: crate::multisig_state::MultisigStateModule {
         self.perform_action(action_id)
     }
 
+    #[event("actionPerformed")]
+    fn perform_action_event(
+        &self,
+        data: ActionFullInfo<Self::Api>,
+    );
```
andrei-marinica, line 94: Serializing everything into a single object makes it hard to decode.
There should be separate events for each action type, especially for async calls. 
The action id is not relevant, you already have it in the call. You should focus on _what_ the action does.

andrei-marinica (changes requested):

andrei-marinica:
Some logging was implemented in multisig, both on each action in general, and on specific action types.

PR #814, opened 2022-08-26
add control changes function in system sc proxy

SkyzoxRobin:
The function "controlChanges" is missing in the system sc proxy


SkyzoxRobin:
I will update with the role "canTransferNFTCreateRole" when the PR [812](https://github.com/ElrondNetwork/elrond-wasm-rs/pull/812) is merged.

SkyzoxRobin:
The PR #812 has been merged, so I added the "canTransferNFTCreateRole" in the properties for the `controlChanges`

andrei-marinica:
The implementation looks good, but it is somewhat inconsistent with the rest of the system SC proxy. We use structs instead of lists of property-value pairs everywhere else.

I would create another struct that holds optionals instead of just booleans. Then we could pass that structure to controlChanges. To avoid code duplication, we can also integrate it everywhere else in the proxy as an intermediary step.

Would you like to do it, or should I take it over? It is doesn't take me long

PR #816, opened 2022-08-31
add create nft with uris in non fungible token mapper

SkyzoxRobin:
The Non-Fungible Token Mapper is great but it lacks a function with the royalties & uris to use for NFT and SFT so I added the function.

I also renamed the `get_all_token_data` to `get_token_data` because I saw some misunderstandings due to the name, some people thought with the name and with the docs that say: _"Gets all the token data for the given nonce. The SC must own at least 1 token to use this function."_ think that you can actually get all the token data for every nonce in the collection if you have at least one token - which is wrong, you meant that when talking about MetaESDT/SFT.

I will also update the documentation with the new functions when this PR is merged.

### DIFF
--- a/contracts/feature-tests/basic-features/src/storage_mapper_non_fungible_token.rs
+++ b/contracts/feature-tests/basic-features/src/storage_mapper_non_fungible_token.rs
@@ -52,6 +52,48 @@ pub trait NonFungibleTokenMapperFeatures:
             .nft_create_and_send(&to, amount, &attributes)
     }
 
+    #[endpoint()]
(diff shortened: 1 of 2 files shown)


dorin-iancu:
If any additional arguments are needed for create, you can always use the nft create function from `self.send()`. I'd much rather not pollute the mapper with every single use-case, otherwise, you'll soon need a function that creates name & with uri, and so on.

The primary design for `FungibleTokenMapper` and `NonFungibleTokenMapper` is for tokens that are used mainly for information through attributes. NFTs with mostly display purposes are out of scope.

PR #822, opened 2022-09-14
implement FromIterator trait for ManagedVec

yankydoo:
With this PR, devs can use the more ergonomic collect fn instead of manually doing a for_each or for loop over the filtered/mapped iterator

Before:
```rust
let mut collected_vec: ManagedVec<i32> = ManagedVec::new();
managed_vec.iter().filter(|x| x <= &5).for_each(|el| collected_vec.push(el));
```
After:

```rust
let collected_vec = managed_vec.iter().filter(|x| x <= &5).collect::<ManagedVec<i32>>();
```


andrei-marinica (approved):

andrei-marinica:
Please sign your commits next time!

PR #833, opened 2022-10-03
implement FromIterator trait for MultiValueEncoded, implement Debug +…

yankydoo:
I implemented FromIterator trait for MultiValueEncoded and also the Debug + PartialEq Traits for MultiValue1 - MultiValue11

I couldn't implement them for MultiValue12+ as Rust tuples with size 12+ don't implement Debug + PartialEq traits (see: https://doc.rust-lang.org/std/primitive.tuple.html#trait-implementations-1)

I'm open to changing the code if you can think of a more elegant solution.

I think it is of value for developers to be able to easily check equality of MultiValues (even if it's just for length up until 11).


andrei-marinica (approved):

andrei-marinica:
Cool implementation, thanks for the contribution!

PR #837, opened 2022-10-13
clippy fixes

andrei-marinica:


dorin-iancu (approved):

CostinCarabas (dismissed):

elrond-codec/src/impl_for_types/impl_bool.rs
```diff
@@ -9,7 +9,9 @@ impl TopEncodeNoErr for bool {
     fn top_encode_no_err<O: TopEncodeOutput>(&self, output: O) {
         // only using signed because this one is implemented in Arwen, unsigned is not
         // TODO: change to set_u64
```
ovstinga, line 11: Can we remove/fix this todo while we make changes here?

ovstinga (dismissed):

elrond-codec/src/impl_for_types/impl_bool.rs
andrei-marinica: there will be refactoring done here, it is not the final form
not yet, though

ovstinga (approved):

elrond-codec/src/impl_for_types/impl_bool.rs
andrei-marinica: there are some subtle implications involving some VM APIs, might go over them after the VM refactor

PR #861, opened 2022-11-18
Multi contract final 2

andrei-marinica:


andrei-marinica:
Was only fixing a bad rebase.

PR #865, opened 2022-11-24
Generalize rewards distribution claim

claudiu725:
- generalized the rewards distribution smart contract in order to allow the user to claim any kind of reward token
- split the rewards into buckets with the `reward_id` key
- the raffle reward is now `nft_reward_percent` which determines the percentage of rewards that a NFT nonce gets from a given raffle regardless of the reward token used
- claiming a the rewards now sets the `was_claimed` flag, which is used to prevent claiming twice


contracts/examples/rewards-distribution/src/rewards_distribution.rs
```diff
@@ -226,32 +237,140 @@ pub trait RewardsDistribution:
...
+                        reward_token_id,
+                        reward_token_nonce,
+                        nft.token_nonce,
+                    )
+                }
+            }
+        }
         if total > 0 {
-            self.send().direct_egld(&caller, &total);
+            self.send()
```
dorin-iancu, line 289: Should probably just return this payment, and accumulate them all in a ManagedVec in the caller function, to then use `direct_multi`.

And of course, separate "egld_amount` or something, since that wouldn't work otherwise.

contracts/examples/rewards-distribution/src/rewards_distribution.rs
```diff
@@ -226,32 +237,140 @@ pub trait RewardsDistribution:
...
+
+        for raffle_id in raffle_id_start..=raffle_id_end {
+            for nft in nfts {
+                if self.try_claim(raffle_id, reward_token_id, reward_token_nonce, &nft) {
+                    total += self.compute_claimable_amount(
+                        raffle_id,
+                        reward_token_id,
+                        reward_token_nonce,
+                        nft.token_nonce,
+                    )
```
dorin-iancu, line 284:
```suggestion
                    );
```

contracts/examples/rewards-distribution/src/rewards_distribution.rs
```diff
@@ -226,32 +237,140 @@ pub trait RewardsDistribution:
...
-        self.send().direct_multi(&caller, &nfts);
     }
 
+    fn try_claim(
+        &self,
+        raffle_id: u64,
+        reward_token_id: &EgldOrEsdtTokenIdentifier,
+        reward_token_nonce: u64,
+        nft: &EsdtTokenPayment,
+    ) -> bool {
```
dorin-iancu, line 300: I suggest using `Result` enum instead of boolean for this. Boolean can be a bit unclear.

contracts/examples/rewards-distribution/src/rewards_distribution.rs
dorin-iancu, line 284: Purely for code indentation, it's usually better to practice early exit conditions.
```rust
        for raffle_id in raffle_id_start..=raffle_id_end {
            for nft in nfts {
                  let claim_result = self.try_claim(...);
                  if claim_result.is_err() { 
                          continue; 
                  }
                
                  // do the current logic here
```

contracts/examples/rewards-distribution/src/rewards_distribution.rs
claudiu725: Done.

contracts/examples/rewards-distribution/src/rewards_distribution.rs
claudiu725: Done.

dorin-iancu (approved):

PR #878, opened 2022-12-08
set token id if empty

dorin-iancu:


claudiu725 (approved):

andrei-marinica (approved):

andrei-marinica:
A test would have been nice, but it's fine for now.

Issue #886, opened 2022-12-09
Allow ContractBase to be specified in contracts and modules traits


gfusee:
Hi! Allowing to specify ContractBase can be very convenient for those who don't use VSCode & Elrond IDE plugin. This enable auto-completion and make IDEs smarter since after macro processing the trait is a ContractBase.

I purpose to modify derives macros to not throw errors when `elrond_wasm::contract_base::ContractBase` is specified in a trait by user. This will make the following code valid : 

```rust
#[elrond_wasm::contract]
pub trait MyContract : elrond_wasm::contract_base::ContractBase + path::to::MyModule {
    // contract code
}
```

```rust
#[elrond_wasm::module]
pub trait MyModule : elrond_wasm::contract_base::ContractBase {
    // module code
}
```

I can make the PR if it can help you.

PR #887, opened 2022-12-10
Allow ContractBase to be specified in contracts and modules traits

gfusee:
I'm coming back about the issue #886.

### What is this pull request about ?

This pull request changes the way macros works. Developers can now explicit write`elrond_wasm::contract_base::ContractBase` as a supertrait of contract/modules traits. It allows IDEs to be way smarter by enabling autocompletion.

### Code changes

When macros process a contract/module trait, I simply filtered supertraits to remove elrond_wasm::contract_base::ContractBase

### Testing

I added a dead simple test : `feature-tests/contract-base-tester`. This test is the adder example contract rewritten in modules with and without `elrond_wasm::contract_base::ContractBase`, mandos tests are used without any change.


Cargo.toml
```diff
@@ -117,6 +117,8 @@ members = [
   "contracts/feature-tests/composability/transfer-role-features/meta",
   "contracts/feature-tests/composability/vault",
   "contracts/feature-tests/composability/vault/meta",
+  "contracts/feature-tests/contract-base-tester",
```
andrei-marinica, line 120: Having an entire new contract just for this feature is a bit much.

You can use `use-module`, since it is also designed to test modules.

We should check that explicit ContractBase is allowed in both module and main contract.

elrond-wasm-derive/src/parse/supertrait_parse.rs
```diff
@@ -1,5 +1,18 @@
 use crate::{model::Supertrait, parse::split_path_last};
 
+pub fn is_contract_base(supertrait: &syn::TypeParamBound) -> bool {
+    match supertrait {
+        syn::TypeParamBound::Trait(t) => {
+            if let Ok(path) = syn::parse_str::<syn::Path>("elrond_wasm::contract_base::ContractBase") {
```
andrei-marinica, line 6: We can do better than this, this limits how `ContractBase` can be specified. It is possible to break the Path into components and only look at the last component.

We are already doing this Path parsing somewhere. This condition can be expressed as "has a module with name ContractBase".

contracts/feature-tests/contract-base-tester/mandos/adder.scen.json
```diff
@@ -0,0 +1,99 @@
+{
```
andrei-marinica, line 1: You don't actually need additional tests for this task, if anything is wrong it will fail at compilation.

andrei-marinica (changes requested):

contracts/feature-tests/contract-base-tester/mandos/adder.scen.json
gfusee: You're right! Done

elrond-wasm-derive/src/parse/supertrait_parse.rs
gfusee: Good point I totally forgot it! Done

andrei-marinica (dismissed):

gfusee:
I saw I need to sign my commits, I will do it

gfusee:
I re-created a PR here : #894. This time with signed commits. I close this PR.

PR #889, opened 2022-12-11
builtin fuctions polymorphic map

andrei-marinica:
An old-school OOP polymorphic map is more flexible than just a match and some functions.

### DIFF
--- a/elrond-wasm-debug/src/tx_execution/builtin_function_mocks/builtin_func_map.rs
+++ b/elrond-wasm-debug/src/tx_execution/builtin_function_mocks/builtin_func_map.rs
@@ -1,0 +1,35 @@
+use std::collections::HashMap;
+
+use crate::tx_mock::TxFunctionName;
+
+use super::builtin_func_trait::BuiltinFunction;
+
+pub struct BuiltinFunctionMap {
+    func_map: HashMap<String, Box<dyn BuiltinFunction>>,
+}
+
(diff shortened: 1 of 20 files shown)


elrond-wasm-debug/src/tx_execution/builtin_function_mocks/builtin_func_map.rs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    pub fn get(&self, name: &TxFunctionName) -> Option<&Box<dyn BuiltinFunction>> {
+        self.func_map.get(name.as_str())
+    }
+}
+
+impl std::fmt::Debug for BuiltinFunctionMap {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("BuiltinFunctionMap").finish()
```
claudiu725, line 33: The `func_map` field has to be added here (eg. `.field("func_map", &self.func_map)`)

claudiu725 (approved):

elrond-wasm-debug/src/tx_execution/builtin_function_mocks/builtin_func_map.rs
claudiu725: I'll add this in a follow-up PR.

PR #892, opened 2022-12-12
Callback payments in debugger

andrei-marinica:


elrond-wasm-debug/src/tx_mock/tx_async_call_data.rs
```diff
@@ -60,10 +69,32 @@ pub fn async_callback_tx_input(async_data: &AsyncCallTxData, async_result: &TxRe
...
+            if !token_transfers.is_empty() {
+                callback_payments.esdt_values = token_transfers.transfers;
+            } else {
+                callback_payments.egld_value = async_call.call_value.clone();
+            }
+            break;
+        }
+    }
+    callback_payments
+}
```
claudiu725, line 96: The payment returned is the first which matches the destination address.
Given the scenario:
`A ---calls---> B ---callback(with payment X)---> A ---calls---> B ---callback(with payment Y)---> A`
Wouldn't the returned payment be `X` in both cases (instead of `X` then `Y`) since `all_calls` isn't cleared?

claudiu725: TODO - will handle in a future PR

claudiu725 (approved):

elrond-wasm-debug/src/tx_mock/tx_async_call_data.rs
andrei-marinica: I think this is the same behavior of the VM. We need to investigate by comparing the two.

PR #899, opened 2022-12-14
Add Fractional NFTs example smart contract

claudiu725:


contracts/examples/fractional-nfts/src/fractional_uri_info.rs
```diff
@@ -0,0 +1,35 @@
...
+        initial_fractional_amount: BigUint<M>,
+    ) -> Self {
+        Self {
+            original_payment,
+            initial_fractional_amount,
+        }
+    }
+
+    pub fn from_uris(uris: ManagedVec<M, ManagedBuffer<M>>) -> Self {
+        let first_uri = uris.get(0);
```
andrei-marinica, line 24: I would add some more explicit error handling here, otherwise the error messages might be hard to understand.

`uris.try_get(0).unwrap_or_else(|| sc_panic!(...))`

contracts/examples/fractional-nfts/src/fractional_uri_info.rs
```diff
@@ -0,0 +1,35 @@
...
+        let serializer = ManagedSerializer::new();
+        serializer.top_decode_from_managed_buffer(&first_uri)
```
andrei-marinica, line 26: Here there are many ways, but maybe the easiest is to make a version of `top_decode_from_managed_buffer` that also accepts a custom base message.

contracts/examples/fractional-nfts/src/fractional_uri_info.rs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    pub fn to_uris(&self) -> ManagedVec<M, ManagedBuffer<M>> {
+        let first_uri = ManagedSerializer::new().top_encode_to_managed_buffer(&self);
+        let mut uris = ManagedVec::new();
```
andrei-marinica, line 31: `ManagedVec::from_single_item(first_uri)`

andrei-marinica (changes requested):

contracts/examples/fractional-nfts/src/fractional_uri_info.rs
claudiu725: Done. I had to use `M::error_api_impl().signal_error` instead of `sc_panic!` since the latter uses `Self::Api` which is available only in the main contract.

contracts/examples/fractional-nfts/src/fractional_uri_info.rs
claudiu725: Added `top_decode_from_managed_buffer_custom_message`.

contracts/examples/fractional-nfts/src/fractional_uri_info.rs
claudiu725: Done.

elrond-wasm/src/contract_base/wrappers/serializer.rs
```diff
@@ -54,6 +54,18 @@ where
         value
     }
 
+    pub fn top_decode_from_managed_buffer_custom_message<T: TopDecode>(
```
andrei-marinica, line 57: Make `top_decode_from_managed_buffer` call this, there is a little duplicate code there.

Also make it take a &'static str, I think it would be nice to move in that direction. It makes code nicer to read and to debug.

andrei-marinica (changes requested):

PR #905, opened 2022-12-16
fix typo

SkyzoxRobin:


andrei-marinica:
Saw there is a conflict, so for simplicity fixed the typo in https://github.com/ElrondNetwork/elrond-wasm-rs/pull/911

PR #932, opened 2023-01-12
StepHandler trait to abstract away creating&executing scenarios

andrei-marinica:
1. Nice trait, but more importantly:
2. Renaming `mandos_*` -> `*_step`


vm/src/mandos_system/model/step_handler.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    #[deprecated(since = "0.39.0", note = "Renamed, use `mandos_set_state` instead.")]
+    fn mandos_set_state(&mut self, step: SetStateStep) -> &mut Self {
+        self.set_state_step(step)
+    }
+
+    #[deprecated(since = "0.39.0", note = "Renamed, use `run_go` instead.")]
+    fn mandos_sc_call(&mut self, step: ScCallStep) -> &mut Self {
+        self.sc_call_step(step)
+    }
```
claudiu725, line 43:
```suggestion
    #[deprecated(since = "0.39.0", note = "Renamed, use `sc_call_step` instead.")]
    fn mandos_sc_call(&mut self, step: ScCallStep) -> &mut Self {
        self.sc_call_step(step)
    }
```

claudiu725 (approved):

PR #936, opened 2023-01-14
SC upgrade tool

andrei-marinica:
Smart contract upgrade tool that can upgrade contracts up to version 0.39.0 (for now).

It is easy to expand with code for new versions. It also has a default implementation that simply bumps versions in the crate configs.

The upgrade tool is part of the meta crate, which can now also be called as a standalone application. 

Also refactored the meta crate CLI arguments using `clap`. Aside from the more concise implementation, we also have the following help:

```
MultiversX smart contract meta-programming tools and build system

Usage: sc-meta [OPTIONS] [COMMAND]

Commands:
  build      Builds contract(s) for deploy on the blockchain.
  build-dbg  Builds contract(s) with symbols and WAT.
  twiggy     Builds contract(s) and generate twiggy reports.
  clean      Clean the Rust project and the output folder.
  snippets   Generates a snippets project, based on the contract ABI.
  upgrade    Upgrades a contract to the latest version. Multiple contract crates are allowed.
  help       Print this message or the help of the given subcommand(s)

Options:
      --no-abi-git-version  Skips loading the Git version into the ABI
  -h, --help                Print help
  -V, --version             Print version


The MultiversX smart contract Meta crate can be used in two ways:
    A. Import it into a contract's specific meta-crate. 
       There it will receive access to the contract ABI generator. 
       Based on that it is able to build the contract and apply various tools.
       This part also contains the multi-contract config infrastructure.

    B. Use it as a standalone tool.
       It can be used to automatically upgrade contracts from one version to the next.
```

Running the upgrade tool produces a colorful output:

Screenshot from 2023-01-17 18-06-20


framework/meta/src/meta_cli.rs
```diff
@@ -0,0 +1,83 @@
...
+use crate::{
+    cli_args::{CliAction, CliArgs},
+    sc_upgrade::upgrade_sc,
+};
+use clap::Parser;
+use multiversx_sc::{
+    abi::ContractAbi, api::uncallable::UncallableApi, contract_base::ContractAbiProvider,
+};
+
+/// The ABI provider set when
```
dorin-iancu, line 11: This comment makes me feel like some words are missing after it.

framework/meta/src/meta_cli.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+fn process_abi<AbiObj: ContractAbiProvider>(cli_args: &CliArgs) -> Option<MetaConfig> {
+    let input_abi = <AbiObj as ContractAbiProvider>::abi();
+    if input_abi.name == NoAbiProvider::NAME {
+        return None;
+    }
+
+    super::meta_validate_abi::validate_abi(&input_abi).unwrap();
```
dorin-iancu, line 35: Any particular reason why you're not importing the function instead of doing the fully qualified call?

dorin-iancu: Also, is `unwrap` instead of `?` intentional?

framework/meta/src/meta_cli.rs
```diff
@@ -0,0 +1,83 @@
...
+                .expect("cannot call `build-dbg` in the standalone meta tool")
+                .build(build_args.into_dbg()),
+            CliAction::Twiggy(build_args) => meta_config_opt
+                .expect("cannot call `twiggy` in the standalone meta tool")
+                .build(build_args.into_twiggy()),
+            CliAction::Clean => meta_config_opt
+                .expect("cannot call `clean` in the standalone meta tool")
+                .clean(),
+            CliAction::GenerateSnippets(gs_args) => meta_config_opt
+                .expect("cannot call snippets in the standalone meta tool")
```
dorin-iancu, line 66: For consistency:
```suggestion
                .expect("cannot call `snippets` in the standalone meta tool")
```

framework/meta/src/meta_cli.rs
```diff
@@ -0,0 +1,83 @@
...
+pub fn cli_main_standalone() {
+    cli_main::<NoAbiProvider>();
+}
+
+pub fn cli_main<AbiObj: ContractAbiProvider>() {
+    let cli_args = CliArgs::parse();
+
+    let meta_config_opt = process_abi::<AbiObj>(&cli_args);
+
+    if let Some(command) = cli_args.command {
```
dorin-iancu, line 51: I feel like this can be rewritten to look slightly better. Normally, I'm also a fan of the `let Some() = variable` patter, but not when it leads to very nested code like this.

I suggest something like:
```
if cli_args.command.is_none() { return; }

let command = unsafe { cli_args.command.unwrap_unchecked() };
if meta_config_opt.is_none()  {
       match command {
CliAction::Upgrade(args) => {
                upgrade_sc(&args);
            },
          _ => // error

        return;
}

let meta_config = unsafe { meta_config_opt.unwrap_unchecked() };
// your match here, but without any "expect"
```

Sorry for bad formatting, blame github. Also, if you want a better formatted message, you can have a method for `CliAction` which gives you the command name as string, i.e. `build`, `build-dbg`, etc.

framework/meta/src/meta_cli.rs
```diff
@@ -0,0 +1,83 @@
...
+            },
+        }
+    }
+}
+
+pub fn multi_contract_config<AbiObj: ContractAbiProvider>(
+    multi_contract_config_toml_path: &str,
+) -> OutputContractConfig {
+    let original_contract_abi = <AbiObj as ContractAbiProvider>::abi();
+    super::meta_validate_abi::validate_abi(&original_contract_abi).unwrap();
```
dorin-iancu, line 79: Same question as before, but why not custom error message (since you can't use `?` operator here)

framework/meta/src/sc_upgrade/folder_structure.rs
```diff
@@ -0,0 +1,95 @@
...
+            dir_type,
+        });
+    }
+}
+
+fn is_marked_contract_crate_dir(path: &Path) -> bool {
+    path.join("multiversx.json").is_file() || path.join("elrond.json").is_file()
+}
+
+fn continue_recursion(dir_entry: &DirEntry) -> bool {
```
dorin-iancu, line 54: Rename to `can_continue_recursion`

framework/meta/src/sc_upgrade/folder_structure.rs
```diff
@@ -0,0 +1,95 @@
...
+    if !dir_entry.file_type().unwrap().is_dir() {
+        return false;
+    }
+
+    if let Some(dir_name_str) = dir_entry.file_name().to_str() {
+        // do not explore hidden folders
+        !dir_name_str.starts_with('.')
+    } else {
+        false
+    }
```
dorin-iancu, line 64: Use `match` instead of let Some() + else

framework/meta/src/sc_upgrade/upgrade_common.rs
```diff
@@ -0,0 +1,166 @@
...
+                            "Renaming {} -> {}",
+                            path.display().to_string().red().strikethrough(),
+                            replaced_path.as_path().display().to_string().green(),
+                        );
+                        fs::rename(path, replaced_path).expect("failed to rename file");
+                    }
+                }
+            }
+        }
+    }
```
dorin-iancu, line 44: Less nesting please. The body of the `if`s should each be a separate function, using early exit conditions instead of deeply nested `if` statements.

framework/meta/src/sc_upgrade/upgrade_common.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    if path.is_dir() {
+        let read_dir = fs::read_dir(path).expect("error reading directory");
+        for child_result in read_dir {
+            let child = child_result.unwrap();
+            version_bump_in_cargo_toml(child.path().as_path(), from_version, to_version);
+        }
+    }
```
dorin-iancu, line 85: Same as above.

framework/meta/src/sc_upgrade/upgrade_common.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+fn upgrade_dependencies_version(
+    cargo_toml_contents: &mut CargoTomlContents,
+    deps_name: &str,
+    from_version: &str,
+    to_version: &str,
+) {
+    if let Some(deps) = cargo_toml_contents.toml_value.get_mut(deps_name) {
```
dorin-iancu, line 94: same

framework/meta/src/meta_cli.rs
andrei-marinica: Changed to function import.

We are pretty top-level at this point, I felt thee was no need to pass errors even higher up.

framework/meta/src/meta_cli.rs
andrei-marinica: In the end got refactored out.

framework/meta/src/meta_cli.rs
andrei-marinica: I split the CLI args in two and removed the Option. It is cleaner now.

framework/meta/src/sc_upgrade/folder_structure.rs
andrei-marinica: Clippy complains about matches with only one destructive pattern, hence the `Option`s always get `if let`

framework/meta/src/sc_upgrade/upgrade_common.rs
andrei-marinica: Fixed by creating smaller functions. Nice catch.

framework/meta/src/sc_upgrade/upgrade_common.rs
andrei-marinica: Also fixed by creating smaller functions. Nice catch.

framework/meta/src/sc_upgrade/upgrade_common.rs
andrei-marinica: Also fixed by creating smaller functions.

dorin-iancu (approved):

alyn509 (approved):

PR #938, opened 2023-01-17
Rust snippets generator fixes

dorin-iancu:
- fix ESDT payments token ID conversion
- generate types for both input and output instead of placeholders (Exception: When a custom user-defined type is an argument, as we'd have no idea how to initialize it, and forcing `Default` implementation doesn't sound like a good idea)
- generate default input values, for easy replacement with custom values


framework/meta/src/generate_snippets/snippet_crate_gen.rs
```diff
@@ -69,8 +69,6 @@ path = ".."
 [dependencies.multiversx-sc-snippets]
 version = "0.39.0"
 
-[workspace]
```
andrei-marinica, line 72: The thing is, if you generate the project for the first time and don't make it a contained workspace, you also have too add the path to the root Cargo.toml.

If the dev wants to make the project permanent it better to have it compile every time, but if he only wants to play around it's better to have it separate ....

Not sure which one is better.

dorin-iancu: Added it back.

claudiu725 (approved):

contracts/feature-tests/rust-snippets-generator-test/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.0.0"
+authors = ["Dorin Marian Iancu <dorin.iancu@multiversx.com>"]
+edition = "2021"
+publish = false
+
+[lib]
+path = "src/lib.rs"
+
+[dependencies.multiversx-sc]
+version = "0.39.2"
```
andrei-marinica, line 12: In the mean time we got to 0.39.4.

framework/meta/src/generate_snippets/snippet_type_map.rs
```diff
@@ -0,0 +1,296 @@
...
+            &self.default_value_expr
+        } else {
+            PLACEHOLDER_INPUT_TYPE_NAME
+        }
+    }
+}
+
+lazy_static! {
+    static ref ABI_TYPES_TO_RUST_TYPES_MAP: HashMap<&'static str, RustTypeString> = {
+        let mut m = HashMap::new();
```
andrei-marinica, line 32: Can we extract this logic into a function?

framework/meta/src/generate_snippets/snippet_type_map.rs
```diff
@@ -0,0 +1,296 @@
...
+
+        m.insert(
```
andrei-marinica, line 34: It might be worth trying to initialize from array for conciseness. Example from official docs:
```
use std::collections::HashMap;

let solar_distance = HashMap::from([
    ("Mercury", 0.4),
    ("Venus", 0.7),
    ("Earth", 1.0),
    ("Mars", 1.5),
]);
```

framework/meta/src/generate_snippets/snippet_type_map.rs
```diff
@@ -0,0 +1,296 @@
...
+    Basic(RustTypeString),
+    Variadic(String),
+    Optional(String),
+    Multi(String),
+    List(String),
+    Array(String, String),
+    Option(String),
+}
+
+/// MultiValueEncoded<BigUint>
```
andrei-marinica, line 150: Leftover comment

andrei-marinica (changes requested):

framework/meta/src/generate_snippets/snippet_type_map.rs
dorin-iancu: Tried it, looks worse due to the fact that we use a struct here.

framework/meta/src/generate_snippets/snippet_type_map.rs
dorin-iancu: Done.

contracts/feature-tests/rust-snippets-generator-test/Cargo.toml
dorin-iancu: Fixed.

framework/meta/src/generate_snippets/snippet_type_map.rs
dorin-iancu: Removed.

andrei-marinica (approved):

PR #951, opened 2023-01-26
Use actions `v2`

claudiu725:
- use `v2` of `contracts.yml` from `mx-sc-actions`
  - this uses `mxpy` instead of `erdpy`, which provides `run-scenarios` instead of `mandos-test`
- updated `multiversx_sc_scenario::run_go` to use `run-scenarios`
  - breaking change note: this means that existing github actions using `contracts.yml@v1` will break and have to be updated to use `v2`
- added handling for the exit status of several more processes, as contract builds were silently failing but the status was reported as successful
- renamed and updated scripts with the following changes:
  - from `elrond.json` to `multiversx.json`
  - from `elrond.workspace.json` to `multiversx.workspace.json`
- renamed some variables


dorin-iancu (dismissed):

ovstinga (approved):

framework/scenario/src/scenario_go_runner.rs
```diff
@@ -10,7 +10,7 @@ pub fn run_go<P: AsRef<Path>>(relative_path: P) {
     let mut absolute_path = std::env::current_dir().unwrap();
     absolute_path.push(relative_path);
 
-    let output = Command::new("mandos-test")
+    let output = Command::new("run-scenarios")
```
andrei-marinica, line 13: Let's make a separate function with this flag, run_go_v2, for backwards compatibility.

I will refactor the whole thing early next week.

andrei-marinica (changes requested):

framework/scenario/src/scenario_go_runner.rs
andrei-marinica: I will also provide a proper test generator for the files where this is called.

andrei-marinica: Or, another option: check that the binary exists before calling, and fallback to the old one if not.

claudiu725: I added a fallback for this, which tries to run `run-scenarios` and if that fails with a `not found` error, it tries the old binary.

andrei-marinica (approved):

andrei-marinica (approved):

andreibancioiu (approved):

ovstinga (approved):

PR #953, opened 2023-01-30
Add attributes for SFT esdts

CostinCarabas:


sdk/core/src/data/esdt.rs
```diff
@@ -1,12 +1,26 @@
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
+use super::address::Address;
+
 // Account holds an Account's information
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct EsdtBalance {
```
dorin-iancu, line 9: Pretty sure this struct was for fungible ESDTs

PR #954, opened 2023-01-30
basic contract template

alyn509:


claudiu725 (approved):

alyn509:
no need of this eny more since we will rename the other contract to fit the convention

PR #955, opened 2023-01-30
Meta crate sources reorg

andrei-marinica:
The crate got past a certain size threshold and needed some modules moved around.


andrei-marinica:
Re-done here: https://github.com/multiversx/mx-sdk-rs/pull/1079

PR #963, opened 2023-02-07
Update gitignore

CostinCarabas:


andrei-marinica:
Doesn't seem relevant anymore. Will re-create if necessary.

PR #973, opened 2023-02-14
Crypto zombies contract

alyn509:
This contract is the result of completing the Crypto Zombies Rust tutorial.


contracts/examples/crypto-zombies/README.md
```diff
@@ -0,0 +1,29 @@
+# Interaction
```
andrei-marinica, line 1: Is this relevant? I don't think we have/want a python interactor for this.

contracts/examples/crypto-zombies/scenarios/adder.scen.json
```diff
@@ -0,0 +1,99 @@
+{
```
andrei-marinica, line 1: If we don't have actual tests yet, better not commit anything here

contracts/examples/crypto-zombies/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+#![no_std]
+
+multiversx_sc::imports!();
+multiversx_sc::derive_imports!();
+
+mod storage;
+mod zombie;
+mod zombieattack;
```
andrei-marinica, line 8: let's name them with underscore, zombie_attack, same for all

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #974, opened 2023-02-15
reconstructor

alyn509:


framework/scenario/src/facade/scenario_world.rs
```diff
@@ -52,15 +52,18 @@ impl ScenarioWorld {
     }
 
     pub fn interpreter_context(&self) -> InterpreterContext {
-        InterpreterContext::new(self.current_dir.clone())
+        InterpreterContext::default()
```
andrei-marinica, line 55: add back the current dir, InterpreterContext::new().with_dir ....

andrei-marinica (changes requested):

framework/scenario/src/scenario/parse_util.rs
```diff
@@ -16,7 +18,8 @@ pub fn parse_scenario_raw<P: AsRef<Path>>(path: P) -> ScenarioRaw {
 
 pub fn parse_scenario<P: AsRef<Path>>(path: P) -> Scenario {
     let scenario_parent = path.as_ref().parent().unwrap();
-    let interpreter_context = InterpreterContext::new(scenario_parent.into());
+    let interpreter_context =
```
andrei-marinica, line 21: Get rid of the `new` with arguments.
Leave a `new` that is the same as the default.
`InterpreterContext::new().with_dir(scenario_parent.into())`

framework/scenario/src/whitebox/contract_obj_wrapper.rs
```diff
@@ -240,6 +241,7 @@ impl BlockchainStateWrapper {
             &wasm_full_path_as_expr,
             &crate::scenario_format::interpret_trait::InterpreterContext::new(
```
andrei-marinica, line 242: InterpreterContext::default()

sdk/scenario-format/src/interpret_trait.rs
```diff
@@ -1,13 +1,26 @@
...
+#[derive(Default, Clone)]
 pub struct InterpreterContext {
     pub context_path: PathBuf,
+    pub vm_type: VMIdentifier,
 }
 
 impl InterpreterContext {
-    pub fn new(context_path: PathBuf) -> Self {
-        InterpreterContext { context_path }
+    pub fn new(context_path: PathBuf, vm_type: VMIdentifier) -> Self {
```
andrei-marinica, line 12: Make it identical to default, no args

sdk/scenario-format/src/value_interpreter/functions.rs
```diff
@@ -1,7 +1,14 @@
+use crate::value_interpreter::*;
 use bech32::FromBase32;
 use sha3::{Digest, Keccak256};
 
-const SC_ADDRESS_NUM_LEADING_ZEROS: usize = 8;
+// SC_ADDRESS_NUM_LEADING_ZEROS is the number of zero bytes every smart contract address begins with.
```
andrei-marinica, line 5: Do not repeat the const/fn name in doc comments. This is only a thing in Go.

sdk/scenario-format/src/value_interpreter/functions.rs
```diff
@@ -1,7 +1,14 @@
...
+pub const SC_ADDRESS_NUM_LEADING_ZEROS: usize = 8;
+
+// SC_ADDRESS_RESERVED_PREFIX_LENGTH is the number of zero bytes every smart contract address begins with.
```
andrei-marinica, line 8: Do not repeat the const/fn name in doc comments. This is only a thing in Go.

andrei-marinica (changes requested):

sdk/scenario-format/src/reconstruct_trait.rs
```diff
@@ -0,0 +1,32 @@
+use std::path::PathBuf;
+
+#[derive(Default)]
+pub struct ReconstructorContext {
+    pub context_path: PathBuf,
```
andrei-marinica, line 5: Context path is not used anywhere. Just leave the struct empty.

sdk/scenario-format/src/interpret_trait.rs
```diff
@@ -1,13 +1,23 @@
 use std::path::PathBuf;
 
-#[derive(Default)]
+use crate::value_interpreter::VMIdentifier;
+
+#[derive(Default, Clone)]
 pub struct InterpreterContext {
     pub context_path: PathBuf,
+    pub vm_type: VMIdentifier,
```
andrei-marinica, line 8: If we added this field, we should probably also use it when interpreting addresses.

Just so we don't forget: search for the function where addresses are being built, and place these 2 bytes on positions 9 and 10 in the address.

Can be done in separate PR.

andrei-marinica (approved):

sdk/scenario-format/src/interpret_trait.rs
alyn509: this was already dome when I added the VM bytes where we build the sc address

PR #984, opened 2023-02-21
Creating the template: Download -> Unzip -> Copy

alyn509:


framework/meta/src/template.rs
```diff
@@ -0,0 +1,50 @@
...
+        .unwrap_or_else(|err| {
+            panic!("error canonicalizing input path: {err}",);
+        })
+        .join(&args.name);
+    copy_template_to_location(&args.name, Path::new(&local_path));
+    Ok(())
+}
+
+pub async fn download_binaries() -> Result<(), reqwest::Error> {
+    let response = reqwest::get(REPOSITORY).await?.bytes().await?;
```
CostinCarabas, line 24: Why use `await` two times?

CostinCarabas (approved):

andrei-marinica (approved):

PR #986, opened 2023-02-22
wasm opt-level = 3

andrei-marinica:


andrei-marinica:
This PR was merely an experiment with the optimization configuration, in the quest for a WebAssembly compilation/execution bug.

It is no longer needed.

PR #989, opened 2023-02-23
Interpreter context missing files handling

ovstinga:


sdk/scenario-format/src/interpret_trait.rs
```diff
@@ -19,6 +20,13 @@ impl InterpreterContext {
             ..self
         }
     }
+
+    pub fn with_missing_files(self, allow_missing_files: bool) -> Self {
```
andrei-marinica, line 24: Let's call this allow_missing_files, just to be clear, and no argument.

andrei-marinica (changes requested):

CostinCarabas (approved):

sdk/scenario-format/src/interpret_trait.rs
ovstinga: Done

sdk/scenario-format/src/interpret_trait.rs
```diff
@@ -19,6 +20,13 @@ impl InterpreterContext {
             ..self
         }
     }
+
+    pub fn with_allow_missing_files(self) -> Self {
```
andrei-marinica, line 24: Sounds a bit awkward, `allow_missing_files()` or `with_allowed_missing_files()` maybe sounds better?

andrei-marinica (approved):

sdk/scenario-format/src/interpret_trait.rs
ovstinga: Agree

PR #990, opened 2023-02-23
Composability interactor refactor

CostinCarabas:
Split State logic in state and config

### DIFF
--- a/contracts/feature-tests/composability/interact/.gitignore
+++ b/contracts/feature-tests/composability/interact/.gitignore
@@ -8,3 +8,4 @@
 
 # Temporary storage of deployed contract address, so we can preserve the context between executions.
 composability_address.txt
+state.toml
--- a/contracts/feature-tests/composability/interact/src/composability_interact_config.rs
+++ b/contracts/feature-tests/composability/interact/src/composability_interact_config.rs
@@ -1,0 +1,32 @@
+use serde::Deserialize;
+use std::io::Read;
+
+/// Config file
+const CONFIG_FILE: &str = "config.toml";
+
+/// Multisig Interact configuration
+#[derive(Debug, Deserialize)]
+pub struct Config {
+    gateway: String,
+    pem: String,
+}
+
+impl Config {
+    // Deserializes config from file
+    pub fn load_config() -> Self {
+        let mut file = std::fs::File::open(CONFIG_FILE).unwrap();
+        let mut content = String::new();
+        file.read_to_string(&mut content).unwrap();
+        toml::from_str(&content).unwrap()
+    }
+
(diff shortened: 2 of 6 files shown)


contracts/feature-tests/composability/interact/src/composability_interact_config.rs
```diff
@@ -0,0 +1,32 @@
+use serde::Deserialize;
+use std::io::Read;
+
+/// Config file
+const CONFIG_FILE: &str = "config.toml";
+
+/// Multisig Interact configuration
+#[derive(Debug, Deserialize)]
+pub struct Config {
+    gateway: String,
```
ovstinga, line 10: according to the `config.toml`, there is no `gateway`, the deserialization will not go through

contracts/feature-tests/composability/interact/src/composability_interact_state.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn set_forwarder_raw_address(&mut self, address: &str) {
+        self.forwarder_raw_address = Some(String::from(address));
+    }
+    /// Sets the promises address
+    pub fn set_promises_address(&mut self, address: &str) {
+        self.promises_address = Some(String::from(address));
+    }
+
+    /// Returns the vault contract
+    pub fn _vault(&self) -> VaultContract {
```
ovstinga, line 54: no need to prefix method name with '_'

CostinCarabas: I don't use this method and I used `_` to silence clippy. Will remove it when I call the method.

contracts/feature-tests/composability/interact/src/composability_interact_config.rs
CostinCarabas: Fixed, thanks!

contracts/feature-tests/composability/interact/.gitignore
```diff
@@ -8,3 +8,4 @@
 
 # Temporary storage of deployed contract address, so we can preserve the context between executions.
 composability_address.txt
```
ovstinga, line 10: old way of storing can be removed

ovstinga (approved):

contracts/feature-tests/composability/interact/src/composability_interact_state.rs
andrei-marinica: You can also do `#[allow(dead_code)]`, gets rid of all warnings

andrei-marinica (approved):

alyn509 (approved):

PR #991, opened 2023-02-24
Error handling for InteractorResult

ovstinga:
Proper error handling for `deploy` and `nft-issue`
Refactor `InteractorResult` methods

### DIFF
--- a/framework/snippets/src/interactor_result.rs
+++ b/framework/snippets/src/interactor_result.rs
@@ -13,6 +13,7 @@ use multiversx_sdk::data::transaction::{
 };
 
 const LOG_IDENTIFIER_SC_DEPLOY: &str = "SCDeploy";
+const LOG_IDENTIFIER_SIGNAL_ERROR: &str = "signalError";
 
 pub struct InteractorResult<T: TopDecodeMulti> {
     pub scrs: Vec<ApiSmartContractResult>,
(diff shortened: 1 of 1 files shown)


framework/snippets/src/interactor_result.rs
```diff
@@ -49,34 +50,70 @@ impl<T: TopDecodeMulti> InteractorResult<T> {
...
+                .as_ref()
+                .unwrap_or_else(|| panic!("missing topics"));
+            assert_eq!(
+                topics.len(),
+                2,
+                "`{LOG_IDENTIFIER_SIGNAL_ERROR}` is expected to have 2 topics"
+            );
+            let error_raw = base64::decode(topics.get(1).unwrap()).unwrap();
+            let error = String::from_utf8(error_raw).unwrap();
+            panic!("error: {error:#?}");
```
andrei-marinica, line 97: In this case we should actually return a Result. In some cases we are explicitly verifying that a tx should fail, and don't want out program to crash.

It's probably best to build some custom error type, or at least an alias for String.

We need:
```
fn handle_signal_error_event(&self) -> Result<(), TxError> ...
pub fn value(&self) -> Result<T, TxError> { ...
pub fn new_deployed_address(&self) ->  Result<Address, TxError>
```
etc.

andrei-marinica (changes requested):

andrei-marinica (approved):

CostinCarabas (approved):

PR #995, opened 2023-03-01
Add scenario for forwarder-queue

CostinCarabas:
Create multiple async calls via forwarder

### DIFF
--- a/contracts/feature-tests/composability/forwarder-queue/src/forwarder_queue.rs
+++ b/contracts/feature-tests/composability/forwarder-queue/src/forwarder_queue.rs
@@ -29,12 +29,6 @@ pub trait ForwarderQueue {
     #[init]
     fn init(&self) {}
 
-    #[proxy]
-    fn self_proxy(&self, to: ManagedAddress) -> crate::Proxy<Self::Api>;
-
-    #[proxy]
-    fn vault_proxy(&self) -> vault::Proxy<Self::Api>;
-
     #[view]
     #[storage_mapper("queued_calls")]
     fn queued_calls(&self) -> LinkedListMapper<QueuedCall<Self::Api>>;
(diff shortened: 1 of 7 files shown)


CostinCarabas:
`run-scenarios "/home/costin/mvx/mx-sdk-rs/contracts/feature-tests/composability/forwarder-queue/scenarios/forw_raw_async_accept_egld.scen.json"`

ovstinga (approved):

contracts/feature-tests/composability/forwarder-queue/src/forwarder_queue.rs
```diff
@@ -64,30 +69,54 @@ pub trait ForwarderQueue {
...
+                    },
+                    QueuedCallType::LegacyAsync => {
+                        contract_call.async_call().call_and_exit();
+                    },
+                    QueuedCallType::TransferExecute => {
+                        contract_call.transfer_execute();
+                    },
+                }
+            } else {
+                let contract_call = self
```
andrei-marinica, line 101: There is no need to work with the proxy, leave this else out and just use the calls above.

We will program it from the data we provide to it, there is no need to for this contract to know about the vault. We pass it the endpoint to call, so we can choose to make it call endpoints from the vault, from itself, or from other contracts.

contracts/feature-tests/composability/forwarder-queue/src/forwarder_queue.rs
```diff
@@ -32,6 +32,9 @@ pub trait ForwarderQueue {
     #[proxy]
     fn self_proxy(&self, to: ManagedAddress) -> crate::Proxy<Self::Api>;
```
andrei-marinica, line 33: No more need for the self proxy, we will allow it to call anything in a raw fashion.

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #997, opened 2023-03-03
Managed Map implementation, tests

andrei-marinica:


contracts/feature-tests/managed-map-features/src/mmap_features.rs
```diff
@@ -0,0 +1,49 @@
+#![no_std]
+
+multiversx_sc::imports!();
+
+#[multiversx_sc::contract]
+pub trait ManagedMapFeatures {
+    #[init]
+    fn init(&self) {}
+
+    #[storage_get("num_entries")]
```
psorinionut, line 10: With the current implementation, we need to keep the number of entries separately, right? Or off-chain.
The same goes for let's say looping (even if it is best avoidable), if I need to loop through the Map, I need to keep a separate list of keys.

andrei-marinica: It's just a test contract, the input is put into storage directly in mandos. The point was to make the test simple and readable.

There is no direct mapping between the ManagedMap and storage, this is just the simplest way to get the data across.

psorinionut: Yes, I saw that the storage is there in order to test the fact that the contract compiles, rather than checking actual data, that is verified in the scenario test. I was referring to something else, so let me rephrase it. Is it too much overhead to have for example the length of the Map stored somehow and retrieve it let's say through a `len()` function? Like with the HashMap? If this is the case, then iterating through the map most probably is a no go.

I am thinking about this as probably most of the SCs that will use a ManagedMap will likely need to know the size of the map. So I wonder which case would be more optimal. Probably, the basic implementation in a SC will be saving a 
ManagedMap in a SingleValueMapper and on top of that save in a new, iterable mapper, the keys of the ManagedMap (that way, you would have access to both the length and the iteration option).

psorinionut (approved):

ovstinga (approved):

PR #1001, opened 2023-03-07
composabiltiy interactor call tree prototype

andrei-marinica:


andrei-marinica:
Recreated here: https://github.com/multiversx/mx-sdk-rs/pull/1008

PR #1002, opened 2023-03-07
Implement `IntoIterator` for storage mappers and some types

### LINKED ISSUE #1031: Collections' `Iter` types are not accessible by SDK users
Collections like `SetMapper`, `MapMapper`, etc. all implement `.iter()` methods, but returned types (`Iter`) are hidden in private modules.

This issue makes it inconvenient to store iterators in user-defined structs, or use iterator types as a concrete type for user-defined traits. One could use `Box<dyn Iteartor>`, but that is inconvenient, inefficient, and can be problematic in `no_std` crates.

olexander-movchan:
I have implemented `IntoIterator` for storage mappers and some other types.

Since iterator types returned by `iter` methods are hidden in private modules, they aren't accessible by SDK users. This issue makes it inconvenient to store iterators in user-defined structs, or use iterator types as a concrete type for user-defined traits. One could use `Box<dyn Iteartor>`, but that is inconvenient and can be problematic in `no_std` crates.

This PR will make it possible to access iterator types like `<&'a MapStorageMapper as IntoIterator>::IntoIter`.

Resolves #1031


andrei-marinica (approved):

andrei-marinica:
Please sign your commits in the future! Learn here how https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account

We'll let it pass this time, but might not in the future.

Other than that, good PR, thank you!

PR #1003, opened 2023-03-07
Update Dependencies

alyn509:
- remove path from `dependencies`
- remove `dev-dependencies`

### DIFF
--- a/framework/meta/src/template.rs
+++ b/framework/meta/src/template.rs
@@ -1,10 +1,13 @@
-use crate::cli_args::TemplateArgs;
+use crate::{cargo_toml_contents::CargoTomlContents, cli_args::TemplateArgs};
 use copy_dir::*;
 use std::{env, fs::File, io::Write, path::Path};
 
 const REPOSITORY: &str = "https://github.com/multiversx/mx-sdk-rs/archive/refs/heads/master.zip";
 const TEMPLATES_SUBDIRECTORY: &str = "mx-sdk-rs-master/contracts/examples/";
 const ZIP_NAME: &str = "./master.zip";
+const ROOT_CARGO_TOML: &str = "./Cargo.toml";
+const META_CARGO_TOML: &str = "./meta/Cargo.toml";
+const WASM_CARGO_TOML: &str = "./wasm/Cargo.toml";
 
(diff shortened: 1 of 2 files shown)


framework/meta/src/template.rs
```diff
@@ -48,3 +57,60 @@ pub fn copy_template_to_location(template: &str, location: &Path) {
...
+    update_dependencies_wasm(template);
+    update_dependencies_meta(template);
+}
+
+pub fn update_dependencies_root(template: &str) {
+    let path_buf = get_canonicalized_path(template, ROOT_CARGO_TOML);
+    let root_cargo_toml_path = Path::new(&path_buf);
+    let mut toml = CargoTomlContents::load_from_file(root_cargo_toml_path);
+
+    toml.dev_dependencies_mut().clear();
```
andrei-marinica, line 72: No, don't clear the dev-dependencies. They are needed for tests. Remove paths like for the regular dependencies.

framework/meta/src/template.rs
```diff
@@ -48,3 +57,60 @@ pub fn copy_template_to_location(template: &str, location: &Path) {
...
+            panic!("error canonicalizing input path: {err}",);
+        })
+}
+
+pub fn remove_paths_from_dependencies(
+    cargo_toml_contents: &mut CargoTomlContents,
+    path: &Path,
+    ignore_deps: &[&str],
+) {
+    let deps_map = cargo_toml_contents.dependencies_mut();
```
andrei-marinica, line 107: You can iterate here through both dependencies and dev-dependencies.

Either chain the iterators, or 2 successive for loops.

andrei-marinica (changes requested):

andrei-marinica (approved):

framework/meta/src/template.rs
```diff
@@ -48,3 +58,63 @@ pub fn copy_template_to_location(template: &str, location: &Path) {
         .join(template);
     let _ = copy_dir(contract_path, location);
 }
+
+pub fn update_dependencies(template: &str) {
```
andrei-marinica, line 62: In future PR: rename to template_crate_name.

We need to explicitly set the new project name too. But not now.

ovstinga (approved):

PR #1007, opened 2023-03-14
comp/interactor+fwd-queue: Remove payment as args

CostinCarabas:
Removed payment from args and provided as a transfer.


CostinCarabas:
Will re-open in another PR.

PR #1016, opened 2023-03-17
Add ticker method to TokenIdentifier

lcswillems:
The `ticker` method returns the ticker of the token identifier.

### DIFF
--- a/framework/base/src/types/managed/wrapped/token_identifier.rs
+++ b/framework/base/src/types/managed/wrapped/token_identifier.rs
@@ -64,6 +64,12 @@ impl<M: ManagedTypeApi> TokenIdentifier<M> {
     pub fn is_valid_esdt_identifier(&self) -> bool {
         M::managed_type_impl().validate_token_identifier(self.buffer.handle.clone())
     }
+
+    pub fn ticker(&self) -> ManagedBuffer<M> {
+        let token_id_len = self.buffer.len();
+        let ticker_len = M::managed_type_impl().get_token_ticker_len(token_id_len);
+        self.buffer.copy_slice(0, ticker_len).unwrap()
+    }
 }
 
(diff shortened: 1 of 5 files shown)


framework/base/src/types/managed/wrapped/token_identifier.rs
```diff
@@ -64,6 +64,12 @@ impl<M: ManagedTypeApi> TokenIdentifier<M> {
     pub fn is_valid_esdt_identifier(&self) -> bool {
         M::managed_type_impl().validate_token_identifier(self.buffer.handle.clone())
     }
+
+    pub fn ticker(&self) -> ManagedBuffer<M> {
```
andrei-marinica, line 68: Please also add some unit tests in `framework/scenario/tests/token_identifier_test.rs`

framework/base/src/types/managed/wrapped/token_identifier.rs
```diff
@@ -64,6 +64,12 @@ impl<M: ManagedTypeApi> TokenIdentifier<M> {
...
+        let token_id_len = self.buffer.len();
+        let ticker_len = M::managed_type_impl().get_token_ticker_len(token_id_len);
+        self.buffer.copy_slice(0, ticker_len).unwrap()
```
andrei-marinica, line 71: We don't like `unwrap()`, because it calls the regular Rust panic, which bloats contracts.

Replace with
`.unwrap_or_else(|_| M::error_api_impl().signal_error(err_msg::BAD_TOKEN_FORMAT))`

And add the error in err_msg.rs like this:
```
...
pub const BAD_TOKEN_PROVIDED: &str = "bad call value token provided";
pub const BAD_TOKEN_FORMAT: &str = "bad token identifier";
...
```
.. or something of the sort

andrei-marinica (changes requested):

PR #1021, opened 2023-03-19
Cleanup multisig/Cargo.toml

andreibancioiu:


contracts/examples/multisig/Cargo.toml
```diff
@@ -20,12 +20,6 @@ path = "../../../contracts/modules"
 version = "0.39.7"
 path = "../../../framework/scenario"
 
-[dev-dependencies.adder]
```
andrei-marinica, line 23: You cannot delete them, we have integrated tests using them.

andrei-marinica (changes requested):

andrei-marinica:
Can ony be done after Wasmer in Rust + new proxies.

PR #1029, opened 2023-03-21
[DO NOT MERGE] simplified factorial example

andrei-marinica:


CostinCarabas (approved):

ovstinga (approved):

andrei-marinica:
The contract gets bigger because of a zero check. Not worth it.

PR #1030, opened 2023-03-22
fwd-queue: Add callback to promises

CostinCarabas:


andrei-marinica (changes requested):

contracts/feature-tests/composability/forwarder-queue/src/forwarder_queue.rs
```diff
@@ -174,12 +175,33 @@ pub trait ForwarderQueue {
...
                 },
                 QueuedCallType::Promise => {
+                    #[cfg(feature = "promises")]
+                    contract_call
+                        .with_gas_limit(call.gas_limit)
+                        .async_call_promise()
+                        .with_callback(self.callbacks().callback_function())
+                        .register_promise();
+
+                    #[cfg(not(feature = "promises"))]
```
CostinCarabas, line 185: Remove this case; it will never be used.

andrei-marinica (approved):

PR #1038, opened 2023-03-30
Release candidate v0.40

andrei-marinica:
All the features for the upcoming v0.40 release.


andrei-marinica:
Two more releases happened in between, so `rc/0.40` becomes `rc/0.42`: https://github.com/multiversx/mx-sdk-rs/pull/1082

PR #1043, opened 2023-04-05
Adder contract interactor

ovstinga:


alyn509 (approved):

contracts/examples/adder/interact/src/adder_interact.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    async fn feed_contract_egld(&mut self) {
+        let _ = self
+            .interactor
+            .transfer(
+                TransferStep::new()
+                    .from(&self.wallet_address)
+                    .to(self.state.adder())
+                    .egld_value("0,050000000000000000"),
```
CostinCarabas, line 151: Not sure, but `0,050000000000000000` value seems odd. I think it should be `50000000000000000`.

CostinCarabas (approved):

contracts/examples/adder/interact/src/adder_interact.rs
andrei-marinica: It's a trick, commas are ignored by the mandos format. It's just hinting at how the value should be read in 18-decimal fixed point.

andrei-marinica (approved):

PR #1052, opened 2023-04-19
Comp interator parallel

CostinCarabas:


contracts/feature-tests/composability/interact/src/call_tree_calling_functions.rs
```diff
@@ -0,0 +1,150 @@
...
+                continue;
+            }
+            println!("successfully performed action 'add_queued_call'");
+        }
+    }
+
+    pub async fn call_root(
+        &mut self,
+        call_state: &CallState,
+        payment_token: EgldOrEsdtTokenIdentifier<DebugApi>,
```
andrei-marinica, line 117: There is no need to transfer ESDT here, the contract will simply forward what it has accumulated

andrei-marinica (changes requested):

contracts/feature-tests/composability/interact/src/call_tree_calling_functions.rs
```diff
@@ -0,0 +1,145 @@
...
+                            .state
+                            .forwarder_queue_from_addr(&fwd_addr_expr)
+                            .add_queued_call(
+                                call_type.clone(),
+                                child_fwd_addr,
+                                DEFAULT_GAS_LIMIT,
+                                FORWARD_QUEUED_CALLS_ENDPOINT,
+                                MultiValueEncoded::<DebugApi, _>::new(),
+                            )
+                            .with_egld_or_single_esdt_transfer(EgldOrEsdtTokenPayment::new(
```
andrei-marinica, line 64: We should go for .with_any_payment, to support multi-ESDT.

Can be done later.

andrei-marinica (approved):

PR #1061, opened 2023-04-25
allocator syntax changes for rustc 1.71.0

andrei-marinica:


andrei-marinica:
Re-created here: https://github.com/multiversx/mx-sdk-rs/pull/1063

PR #1068, opened 2023-04-27
count api calls and provide them in tx_results

yankydoo:
# Count API Calls

## Problem

The [API Limits](https://docs.multiversx.com/developers/contract-api-limits/) are not testable with pure rust tests.

## Solution

I extended the rust test framework to count the API calls:

- [x] Trie Reads
- [x] Transfers
- [ ] Built-in function calls

This enables to write tests that count these calls and make assert statements like these:

```rust
assert_eq!(tx_result.api_calls.trie_reads, 5);
assert_eq!(tx_result.api_calls.transfers, 2);
```

## TODO

I could not find a clear point where to count built-in function calls (for me, it is also unclear what exactly counts as such). Could you please help me and guide me to a solution?

Please also verify I counted the transfers and trie_reads correctly.

Apart from that, I still would like to extend some tests or write new tests to verify the correct counting.

Summarized:

- [ ] Count built-in function calls
- [ ] Verify correctness of trie read counts
- [ ] Verify correctness of transfer counts
- [ ] Extend/write tests

### DIFF
--- a/framework/scenario/src/whitebox/contract_obj_wrapper.rs
+++ b/framework/scenario/src/whitebox/contract_obj_wrapper.rs
@@ -747,9 +747,10 @@ impl BlockchainStateWrapper {
         let exec_result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| tx_fn(sc)));
 
         let api_after_exec = Rc::try_unwrap(TxContextStack::static_pop()).unwrap();
+        let api_calls = (*api_after_exec.api_calls.borrow()).clone();
         let updates = api_after_exec.into_blockchain_updates();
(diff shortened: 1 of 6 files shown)


andrei-marinica:
Trie reads are difficult to count.

First of all, this is a simulator of the VM, so it definitely has no access to actual trie reads. Please also note that the VM caches reads, so that needs to be taken into account.

Such a tool would be more suitable for the VM executor.

In any case, it can also be estimated here, but beware. We are going to undergo massive refactoring of the APIs, and the proper place to add this counts will be in the VM, after splitting.

I will think of what we can do with this, but it is important to get the refactoring done first. It's a pretty heavy operation.

PR #1074, opened 2023-05-03
Fix assert positive

CostinCarabas:


framework/base/src/api/managed_types/big_int_api.rs
```diff
@@ -58,4 +58,6 @@ pub trait BigIntApi: HandleTypeInfo {
     fn bi_shl(&self, dest: Self::BigIntHandle, x: Self::BigIntHandle, bits: usize);
 
     fn bi_to_string(&self, bi_handle: Self::BigIntHandle, str_handle: Self::ManagedBufferHandle);
+
+    fn bi_assert_positive(&self, x: Self::BigIntHandle);
```
andrei-marinica, line 62: Do not expose this in the API, nobody will ever call this directly. Only test the bitwise operations.

andrei-marinica (changes requested):

CostinCarabas:
Re-open issue https://github.com/multiversx/mx-sdk-rs/pull/1075

PR #1086, opened 2023-05-09
update default issue callback

alyn509:


sasurobert (approved):

framework/base/src/storage/mappers/token_mapper_state.rs
```diff
@@ -0,0 +1,28 @@
+use crate::{
+    api::ManagedTypeApi,
+    codec::{
+        self,
+        derive::{NestedDecode, NestedEncode, TopDecode, TopEncode},
+    },
+    types::TokenIdentifier,
+};
+
+#[derive(NestedEncode, NestedDecode, TopEncode, TopDecode, Default)]
```
andrei-marinica, line 10: This encoding is not backwards compatible with the old TokenIdentifier stored in the token mapper.

To preserve backwards compatibility, it is possible to define an encoding where "" means TokenMapperState::NotSet, "pending" or something for Pending, and the verbatim token identifier for TokenIdentifier

andrei-marinica (changes requested):

lcswillems:
Why do we have a `clear` method? If we prevent the token to be issued twice, then having a `clear` method enables to escape this, no? I emit, I clear, I emit again. @sasurobert @andrei-marinica

lcswillems (approved):

andrei-marinica (approved):

PR #1087, opened 2023-05-09
Scenario support for `.mxsc.json` files

CostinCarabas:
The mxsc files are great for packing contract binaries with an ABI and more metadata.

After the release of the new mx-scneario-go CLI, they can now be added to the Rust backend.

Also needed to slightly adjust the templating engine.


sdk/scenario-format/tests/interpreter_test.rs
```diff
@@ -31,6 +31,25 @@ fn test_string() {
...
+#[test]
+fn test_mxsc_pack() {
+    let context = &InterpreterContext::default();
+    assert_eq!(
+        b"test".to_vec(),
+        interpret_string("mxsc:tests/test.json", context)
+    );
+}
+
+#[should_panic]
```
andrei-marinica, line 43: Unintuitively, should not panic.
When running using the debugger, we don't actually rely on these files. Their path is their sole identifier in this case, and only used as an identifier.
`file:` does not panic, neither should this. This fix should solve the failing rust tests in the CI.

andrei-marinica (changes requested):

andrei-marinica (approved):

mihaicalinluca (approved):

BiancaIalangi (approved):

PR #1088, opened 2023-05-09
human readable format tool prototype

andrei-marinica:
Just a prototype, do not merge!

Also note that this is part of a challenge, for now. Do not interfere.


andrei-marinica:
Replaced by https://github.com/multiversx/mx-sdk-rs/pull/1273

PR #1090, opened 2023-05-11
unified ScenarioWorld for all backends

andrei-marinica:
One object to rule them all.

The goal is to create a unified interface for all tests, independent of the "backend" running the test.


alyn509 (approved):

ovstinga (approved):

framework/scenario/src/facade/scenario_world.rs
```diff
@@ -6,34 +6,85 @@ use crate::{
...
+use super::debugger_backend::DebuggerBackend;
+
 /// A facade for contracts tests.
 ///
 /// Contains all the context needed to execute scenarios involving contracts.
 ///
 /// Currently defers most of the operations to the blockchain mock object directly,
 /// but that one will be refactored and broken up into smaller pieces.
-#[derive(Default, Debug)]
+// #[derive(Default, Debug)]
```
andrei-marinica, line 22: Cleaned up in the last PR.

CostinCarabas (approved):

PR #1100, opened 2023-06-05
System SC mock

andrei-marinica:
A system SC mock is required to run tests that involve issuing of new tokens.


vm/Cargo.toml
```diff
@@ -26,6 +26,8 @@ rand = "0.8.5"
 rand_seeder = "0.2.2"
 ed25519-dalek = "1.0.1"
 itertools = "0.10.3"
+bech32 = "0.9"
```
andrei-marinica, line 29: I don't think this dependency is needed.

Bech32 is more of a "front-end" thing.

vm/src/tx_execution/system_sc.rs
```diff
@@ -0,0 +1,55 @@
...
+    address.as_array() == &ESDT_SYSTEM_SC_ADDRESS_ARRAY
+}
+
+pub fn execute_system_sc(tx_context: TxContext) -> (TxContext, TxResult) {
+    let func_name = tx_context.tx_input_box.func_name.clone();
+    match func_name.as_str() {
+        "issue" => issue(tx_context),
+        "issueSemiFungible" => issue_semi_fungible(tx_context),
+        "issueNonFungible" => issue_non_fungible(tx_context),
+        "registerMetaESDT" => todo!(),
```
andrei-marinica, line 24: I would replace these todos with a more explicit panic message

vm/src/tx_execution/system_sc/system_sc_issue.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    if tx_input.args.len() < 2 {
+        tx_result = TxResult::from_vm_error("not enough arguments");
+        return (tx_context, tx_result);
+    }
+    let _name = tx_input.args[0].clone();
+    let ticker = tx_input.args[1].clone();
+
+    let mut new_token_identifiers = tx_cache.get_new_token_identifiers();
+    let token_identifier = if let Some(ti) = new_token_identifiers.pop_front() {
```
andrei-marinica, line 64: It might not be the first token identifier the one that has the right ticker.

We must iterate while extracting the ticker, then remove at that position.

This is also some duplicate code, which we should extract to a separate method in TxCache.

vm/Cargo.toml
ovstinga: Done

vm/src/tx_execution/system_sc.rs
ovstinga: Added blanket `fn`s for now with `unimplemented!()`

vm/src/tx_execution/system_sc/system_sc_issue.rs
ovstinga: Done

ovstinga (approved):

alyn509 (approved):

CostinCarabas (approved):

PR #1101, opened 2023-06-07
bump rand and bip39 in multiverse-sdk

rupansh:
the older version of bip39 was causing version conflicts for unicode-normalize for me
also bump rand version.
Note that i switched to `generate_in` as we were effectively using the same code and bip39 doesn't support rand 0.8


andrei-marinica:
Hi! Could you perform another merge from master?
Thanks!

andrei-marinica (approved):

andrei-marinica:
Hi! Turns out I didn't notice and more commits entered master. A merge is required again.

rupansh:
merged @andrei-marinica

andrei-marinica (approved):

PR #1116, opened 2023-06-16
Off-chain type parsing

gfusee:
# Motivations

I added a way to construct VM query requests and to parse the result by using the same typing system as the WASM framework.

This allows backend developers to work with contract datas in a MUCH convenient way, by easily create VM requests with known types (ManagedBuffer, BigUint, ManagedVec, etc...) and parsing results into "native" rust types (String, num_bigint, Vec, etc...). This is also compatible with user-defined struct, as long as they derive the 4 traits Top/Nested Encode/Decode.

The code I purpose here has been built with a second idea in mind for the future : automatically generate contract interaction crates from a given ABI, like swagger-codegen does.

# How does it work ?

## VM query creation

I implemented a method named `push_managed_arg` on the `VmValueRequest` struct, this method takes a `TopEncode` instance and automatically converts it to a hex formatted argument, compatible with the /vm-values/query gateway's endpoint.

A second method has been implemented, `push_multi_managed_arg`, it does the same thing but for `TopEncodeMulti` instances.

## Parsing results

I added a generic method `get_parsed_return_data` on the VMOutputApi struct, that take an argument's index and converts the base64 string from the /vm-values/query endpoint to a **"native"** rust type. I really wanted to not have a ManagedType result here but a common type more convenient to use in an off-chain context.

I also added `get_parsed_return_data_multi_in_range` and `get_parsed_return_data_var_args` methods to handle multi results.

All of these three methods rely on the dead-easy `NativeConvertible` trait that is responsible to bridge a managed type to a "native" rust one.

I also created two structs, `NativeValue` and `NativeValueManagedVecItem` that act as a wrapper for user based structs, to make them compatible with `NativeConvertible`.

I wrote automated tests for everything, I just have to do some tests on a real environment. I'm looking forward for your thoughts about it :)

### DIFF
--- a/sdk/core/src/data/types/native.rs
+++ b/sdk/core/src/data/types/native.rs
@@ -50,29 +50,6 @@ where
     }
 }
 
-/*impl<M, T> ManagedType<M> for NativeValueManagedVecItem<M, T>
-where
-    M: ManagedTypeApi,
-    T: ManagedType<M> + TopEncode + NestedEncode + TopDecode + NestedDecode + ManagedVecItem + Clone
-{
-    type OwnHandle = T::OwnHandle;
-
-    fn from_handle(handle: Self::OwnHandle) -> Self {
-        NativeValueManagedVecItem::new(T::from_handle(handle))
-    }
-
-    fn get_handle(&self) -> Self::OwnHandle {
-        self.value.get_handle()
-    }
-
-    fn transmute_from_handle_ref(handle_ref: &Self::OwnHandle) -> &Self {
-        unsafe {
-            core::mem::transmute(handle_ref)
-        }
-    }
-}*/
-
-
 impl<T> ManagedVecItem for NativeValueManagedVecItem<T>
 where
     T: TopEncode + NestedEncode + TopDecode + NestedDecode + ManagedVecItem + Clone


sdk/core/src/data/types/buffer.rs
```diff
@@ -0,0 +1,27 @@
+use std::str;
+use multiversx_sc::api::ManagedTypeApi;
+use multiversx_sc::types::{ManagedBuffer, TokenIdentifier};
+use crate::data::types::native::NativeConvertible;
+
+impl<M: ManagedTypeApi> NativeConvertible for ManagedBuffer<M> {
+    type Native = String;
+
+    fn to_native(&self) -> Self::Native {
```
andrei-marinica, line 9: There is some overlap in functionality between this and CodecFrom/CodecInto.

The difference is CodecFrom/CodecInto go many-to-many, whereas here it is many-to-one.

We used to have something similar in the past. Could be easier to use when returning data, since you only have 1 choice of conversion.

Something like this would rather belong in the codec or the base framework.

sdk/core/tests/vmquery_test.rs
```diff
@@ -0,0 +1,612 @@
...
+const SAMPLE_RETURN_DATA: [&'static str; 8] = [
+    "VGhpcyBpcyBhIGJ1ZmZlcg==", // "This is a buffer",
+    "DeC2s6dkAAA=", // 1e18
+    "AAAAAAAAAAAFAPEt0QxNK+gmT+M52hS5+te982SufOs=", // erd1qqqqqqqqqqqqqpgq7ykazrzd905zvnlr88dpfw06677lxe9w0n4suz00uh,
+    "AAAAFFRoaXMgaXMgYSBzYW1wbGUgc3RyAAAAAgPoAAAAAAAAAAAFAPEt0QxNK+gmT+M52hS5+te982SufOs=", // SampleCodable : "This is a sample str", 1000, erd1qqqqqqqqqqqqqpgq7ykazrzd905zvnlr88dpfw06677lxe9w0n4suz00uh
+    "V0VHTEQtYWJjZGVm", // "WEGLD-abcdef"
+    "Zmlyc3Q=", // "first"
+    "c2Vjb25k", // "second"
+    "dGhpcmQ=" // "third"
+];
```
andrei-marinica, line 25: You should take a look at interactors/snippets (preferably in rc/v0.42). We have a full framework for querying the blockchain with SC abstractions and automated data parsing, plus ability to automatically generate integration tests from the interactions.

andrei-marinica (changes requested):

andrei-marinica:
What you want to achieve is very cool, be mindful that we are very soon releasing something in a very similar mindset: blockchain interactors that use the types and code from smart contracts directly.

Have a look at a simple such example:
https://github.com/multiversx/mx-sdk-rs/blob/rc/v0.42/contracts/examples/adder/interact/src/adder_interact.rs

sdk/core/src/data/types/buffer.rs
gfusee: The key advantage of having a many-to-one instead of a many-to-many regarding the output is that the end backend developer doesn't have to think about types. My motivation here is really to hide the "complexity" of what a ManagedType is.

Many-to-many will be useful to do a similar thing about parameter input though

sdk/core/tests/vmquery_test.rs
gfusee: I didn't know! I checked it and this is really great! I'll wait for when you merge it into main before continuing with what I want to achieve

PR #1125, opened 2023-06-20
DebuggerApi - managed map, big float

andrei-marinica:


vm/src/api/managed_types/managed_map_api_mock.rs
```diff
@@ -65,9 +29,19 @@ impl ManagedMapApiImpl for DebugApi {
         key_handle: Self::ManagedBufferHandle,
         value_handle: Self::ManagedBufferHandle,
     ) {
-        let key = self.mb_get(key_handle);
-        let value = self.mb_get(value_handle);
-        self.mm_values_insert(map_handle, key, value);
+        let key = self
+            .m_types_borrow()
+            .mb_get(key_handle.get_raw_handle_unchecked())
```
alyn509, line 34: maybe extract this in a function for a brief simplification

alyn509 (approved):

vm/src/api/managed_types/managed_map_api_mock.rs
andrei-marinica: This code gets deleted a few PRs later. We only need the tests to pass for now.

ovstinga (approved):

PR #1163, opened 2023-07-13
getting smart contract results from logs

gfusee:
Hey!

Sometimes, for a reason I ignore, the smartContractResults field is not present in the proxy's /transactions/:txHash?withResults=true endpoint, despite the transaction actually have results. In this case we can get them from logs.

In order to handle this I changed the way you parse the response.

Just a word the `issue_non_fungible_new_token_identifier` function. The previous signature was `Result<String, TxResponseStatus>`, but the function either finish with `Ok(...)` or never finish (panic via the macro or via forced unwrapping). I might miss something here.
I would have changed the signature to `Option<String>` (and so removing the `Result<...>`), but I saw in the `multisig_interact_nfts.rs` and `interactor_sc_call.rs` files that you are printing the error, so I think you plan to implement it later? So finally I left the `Result` as is and the output is `Result<Option<String>, TxResponseStatus>`

### DIFF
--- a/contracts/examples/multisig/interact-rs/src/multisig_interact_nfts.rs
+++ b/contracts/examples/multisig/interact-rs/src/multisig_interact_nfts.rs
@@ -105,7 +105,7 @@ impl MultisigInteract {
             return;
         }
 
-        self.collection_token_identifier = result.unwrap();
+        self.collection_token_identifier = result.unwrap().unwrap();
         println!(
             "collection token identifier: {}",
             self.collection_token_identifier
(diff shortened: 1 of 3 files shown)


andrei-marinica:
The actual fix was a lot more complicated than that.

Was fixed in #1165

gfusee:
unfortunately this is not fixed yet, #1165 fixes the panic that occurs when smartContractResults

but it doesn't fixes the fact that no smartContractResults != no contract result, in order to get results in this case we must check into logs like I did in this PR

PR #1176, opened 2023-07-24
results from logs

gfusee:
cf my last comment on #1163


andrei-marinica (approved):

andrei-marinica:
It looks ok, but please test it once more, since we don't have CI for this.

gfusee:
I added some tests and made a refacto

Interesting thing : while in sync calls the field smartContractResults might not be present in the transaction infos, it seems to be always there in async calls

andrei-marinica (approved):

PR #1178, opened 2023-07-25
made Interactor thread safe

gfusee:
I created the ScenarioRunnerListSafe struct that is Send + Sync, allowing an Interactor to be Send + Sync too


andrei-marinica:
Is there any reason to also keep the version that does not implement Send + Sync? It sounds like we would always want the safe version.

gfusee:
I don't know what are your plan about that! Personally I don't see any reason to keep the version that doesn't implement Send + Sync for off-chain operation
For on-chain operations it may be worth to keep not Send + Sync since I don't how Wasmer handle atomic operations and at which cost

I just made another PR complementary to this one that implements Send + Sync and use Arc/Mutex to make the scenario engine works on multithreading too : #1182

andrei-marinica:
Yes, the other PR (https://github.com/multiversx/mx-sdk-rs/pull/1182) is the way to go. The wasm API does not need to know about Send+Sync, it's only used in the debugger. The implementation there is good.

PR #1182, opened 2023-08-09
Made ScenarioWorld and related stuff thread safe

gfusee:
I replace non Send and Sync types Rc and RefCell by their equivalent Arc and Mutex
I also added Send + Sync constraints where needed

I changed some names accordingly to the changes made

### DIFF
--- a/framework/scenario/src/facade/scenario_world.rs
+++ b/framework/scenario/src/facade/scenario_world.rs
@@ -145,7 +145,7 @@ impl ScenarioWorld {
         self.get_mut_debugger_backend()
             .vm_runner
             .contract_map_ref
-            .borrow_mut()
+            .borrow()
             .register_contract(contract_bytes, contract_container);
     }
 
--- a/framework/scenario/src/facade/scenario_world.rs
+++ b/framework/scenario/src/facade/scenario_world.rs
@@ -145,7 +145,7 @@ impl ScenarioWorld {
         self.get_mut_debugger_backend()
             .vm_runner
             .contract_map_ref
-            .borrow()
+            .lock()
             .register_contract(contract_bytes, contract_container);
     }
 
--- a/framework/scenario/src/whitebox_legacy/contract_obj_wrapper.rs
+++ b/framework/scenario/src/whitebox_legacy/contract_obj_wrapper.rs
@@ -262,7 +262,7 @@ impl BlockchainStateWrapper {
             .get_mut_debugger_backend()
             .vm_runner
             .contract_map_ref
-            .borrow()
+            .lock()
             .contains_contract(contract_code_expr.value.as_slice());
         if !contains_contract {
             let contract_obj = create_contract_obj_box(obj_builder);
@@ -271,7 +271,7 @@ impl BlockchainStateWrapper {
                 .get_mut_debugger_backend()
                 .vm_runner
                 .contract_map_ref
-                .borrow()
+                .lock()
                 .register_contract(
                     contract_code_expr.value,
                     ContractContainer::new(contract_obj, None, false),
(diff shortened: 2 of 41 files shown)


gfusee:
This is a draft bc I'm gonna write at least one test that ensures that a scenario can be run in a multithreading context (probably using tokio::spawn)

The code itself will not change

andrei-marinica (dismissed):

framework/scenario/src/facade/scenario_world.rs
```diff
@@ -145,7 +145,7 @@ impl ScenarioWorld {
         self.get_mut_debugger_backend()
             .vm_runner
             .contract_map_ref
-            .borrow_mut()
+            .borrow()
```
andrei-marinica, line 148: Why is `borrow_mut` no longer necessary? This should be a mutable operation.

framework/scenario/src/whitebox_legacy/contract_obj_wrapper.rs
```diff
@@ -270,7 +271,7 @@ impl BlockchainStateWrapper {
                 .get_mut_debugger_backend()
                 .vm_runner
                 .contract_map_ref
-                .borrow_mut()
+                .borrow()
```
andrei-marinica, line 274: Why is borrow_mut no longer necessary? This should be a mutable operation.

framework/scenario/tests/contract_without_macros.rs
```diff
@@ -286,7 +286,7 @@ mod sample_adder {
 
     impl<A> multiversx_sc::contract_base::CallableContract for ContractObj<A>
     where
-        A: multiversx_sc::api::VMApi,
+        A: multiversx_sc::api::VMApi + Send + Sync,
```
andrei-marinica, line 289: Ideally, the contract shouldn't know about Send/Sync. Doesn't it work without this?

framework/scenario/tests/contract_without_macros.rs
```diff
@@ -299,7 +299,7 @@ mod sample_adder {
     pub struct ContractBuilder;
 
     impl multiversx_sc::contract_base::CallableContractBuilder for ContractBuilder {
-        fn new_contract_obj<A: multiversx_sc::api::VMApi>(
+        fn new_contract_obj<A: multiversx_sc::api::VMApi + Send + Sync>(
```
andrei-marinica, line 302: Ideally, the contract shouldn't know about Send/Sync. Doesn't it work without this?

andrei-marinica (changes requested):

alyn509 (approved):

framework/scenario/src/facade/scenario_world.rs
gfusee: with Mutex there is only a lock mechanism, no difference between borrow and borrow_mut like there is on RefCell, I'll rename the function to reflect that

framework/scenario/src/whitebox_legacy/contract_obj_wrapper.rs
gfusee: with Mutex there is only a lock mechanism, no difference between borrow and borrow_mut like there is on RefCell, I'll rename the function to reflect that

framework/scenario/tests/contract_without_macros.rs
gfusee: Done by adding the constraint Send + Sync on traits VMApi and VMHooksApiBackend

framework/scenario/src/facade/scenario_world.rs
gfusee: done by renaming the function "borrow" to "lock" to better match the RefCell -> Mutex conversion

"lock" = both old "borrow" and "borrow_mut" functions

framework/scenario/src/whitebox_legacy/contract_obj_wrapper.rs
gfusee: done by renaming the function "borrow" to "lock" to better match the RefCell -> Mutex conversion

"lock" = both old "borrow" and "borrow_mut" functions

andrei-marinica (approved):

PR #1197, opened 2023-08-16
vm: Add get_shard_of_address

CostinCarabas:


vm/src/vm_hooks/vh_handler/vh_blockchain.rs
```diff
@@ -35,6 +35,10 @@ pub trait VMHooksBlockchain: VMHooksHandlerSource {
         );
     }
 
+    fn get_shard_of_address(&self, address_bytes: &[u8]) -> i32 {
+        (address_bytes[address_bytes.len()] % 3).into()
```
andrei-marinica, line 39: `address_bytes.len() - 1`
We need at least one test before merging this.

vm/src/vm_hooks/vh_handler/vh_blockchain.rs
```diff
@@ -35,6 +35,10 @@ pub trait VMHooksBlockchain: VMHooksHandlerSource {
         );
     }
 
+    fn get_shard_of_address(&self, address_bytes: &[u8]) -> i32 {
```
andrei-marinica, line 38: You also need some glue code in vh_dispatcher.rs, I believe that is also missing.

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #1200, opened 2023-08-18
Remove deprecated files

andreibancioiu:


zip-example-wasm.sh
```diff
@@ -1,23 +0,0 @@
-#!/bin/sh
```
andrei-marinica, line 1: I might still need this

andrei-marinica (changes requested):

andrei-marinica:
I fixed it, no need to delete.

PR #1202, opened 2023-08-21
template cargo remove paths that are not framework

alyn509:


framework/meta/src/template/template_adjuster.rs
```diff
@@ -28,8 +29,9 @@ impl TemplateAdjuster {
         let cargo_toml_path = self.target.contract_dir().join(ROOT_CARGO_TOML);
         let mut toml = CargoTomlContents::load_from_file(&cargo_toml_path);
 
+        let non_framework_crate_names = toml.local_dependency_paths(FRAMEWORK_CRATE_NAMES);
```
andrei-marinica, line 32: Create a method that deletes framework paths directly in the toml object.

Also note that local_dependency_paths does not handle dev-dependencies, so make sure to handle both.

andrei-marinica (changes requested):

andrei-marinica:
Currently, the functionality attempted here was achieved in a different way, and is currently released.

It might still be relevant if we decide to have "workspace templates" that include multiple contracts.

However, this doesn't seem to be relevant in the short and medium term.

PR #1210, opened 2023-09-08
send wrapper non zero endpoints

alyn509:


framework/base/src/contract_base/wrappers/send_wrapper.rs
```diff
@@ -222,6 +279,24 @@ where
...
+        &self,
+        to: ManagedAddress<A>,
+        token: TokenIdentifier<A>,
+        nonce: u64,
+        amount: BigUint<A>,
+    ) -> ! {
+        if amount == 0 {
+            ContractCallNoPayment::<A, ()>::new(to, ManagedBuffer::new())
+                .async_call()
+                .call_and_exit_ignore_callback()
```
andrei-marinica, line 292: No, there is no token to be transferred!
```
 pub fn transfer_esdt_non_zero_via_async_call(
        &self,
        to: ManagedAddress<A>,
        token: TokenIdentifier<A>,
        nonce: u64,
        amount: BigUint<A>,
    ) {
        if amount == 0 {
            return;
```

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #1214, opened 2023-09-12
add caller check is depositor on fund

alyn509:


sasurobert (approved):

contracts/examples/digital-cash/src/digital_cash.rs
```diff
@@ -30,6 +30,11 @@ pub trait DigitalCash {
     fn fund(&self, address: ManagedAddress, valability: u64) {
         let deposit_mapper = self.deposit(&address);
         require!(!deposit_mapper.is_empty(), FEES_NOT_COVERED_ERR_MSG);
+        let depositor = deposit_mapper.get().depositor_address;
```
sasurobert, line 33: we have to simplify - take fees in the fund function, should not need to make 2 txs.

sasurobert (changes requested):

contracts/examples/digital-cash/src/digital_cash.rs
alyn509: we cannot make eGLD + ESDT in 1 transaction

sasurobert: but you can take, wEGLD in case of ESDT.

sasurobert (approved):

PR #1216, opened 2023-09-14
Changed signature of sc_deploy_* functions to allow AsMut instances

gfusee:
This change is similar to what's done with `sc_call` function. It allows to be more flexible and it also allow to pass mutable references instead of direct values to the `sc_deploy_*_result` functions


andrei-marinica:
`sc_deploy_get_result` and `sc_deploy_use_result` are now deprecated, replaced by the unified tx syntax, so the changes are no longer needed.

PR #1221, opened 2023-09-19
Digital Cash - allow configure fee token accepted

alyn509:


contracts/examples/digital-cash/wasm/src/lib.rs
```diff
@@ -20,13 +22,14 @@ multiversx_sc_wasm_adapter::endpoints! {
     digital_cash
     (
         init => init
+        claimFees => claim_fees
+        getAmount => get_amount
+        pay_fee_and_fund => pay_fee_and_fund
         fund => fund
+        depositFees => deposit_fees
         withdraw => withdraw
         claim => claim
```
sasurobert, line 31: would like to see set/remove fee token. Also fee tokens should be a map of token/value. And SC owner can change those at any time.

contracts/examples/digital-cash/src/pay_fee_and_fund.rs
```diff
@@ -0,0 +1,43 @@
+multiversx_sc::imports!();
+multiversx_sc::derive_imports!();
+
+use crate::{constants::*, helpers, storage};
+
+#[multiversx_sc::module]
+pub trait PayFeeAndFund: storage::StorageModule + helpers::HelpersModule {
+    #[endpoint]
+    #[payable("*")]
+    fn pay_fee_and_fund(&self, address: ManagedAddress, valability: u64) {
```
sasurobert, line 10: this does not work with eGLD. Add implementation for that.

sasurobert: in case of eGLD you have only 1 payment. so according to tokenID - make different implementaiton.

alyn509: we still have the classic deposit_fees (call 1) -> fund (call 2) 
this was not changed. This new implementation of pay_fee_and_fund comes extra for ESDT fund and fee only

sasurobert (approved):

PR #1227, opened 2023-09-27
NFT Subscription module

thounyy:
This PR adds support for subscription with duration to nfts via a new module. This is an adaptation of the EIP-5643 for MultiversX, more here https://eips.ethereum.org/EIPS/eip-5643.
It is tested in the nft_subscription smart contract with scenarios.


alyn509 (dismissed):

andrei-marinica (dismissed):

andrei-marinica:
Nice work!

However, you have a commit that is not signed. Could you re-create this or force push with all commits signed?

Also would be nice run a `cargo fmt` at the end. The rust tests should also be capable of running, we can help you with that.

thounyy:
just signed all commits and ran cargo fmt as well. Is there anything else I should do?

nvergez:
Hello,

Don't you think it's a problem to offer a module that uses NFT attributes exclusively to track expiration? What happens if I mint an NFT with attributes that have nothing to do with a u64 representing the expiration and then call the renew_subscription method?

I think if a developer uses this module, they should be aware that they must use on-chain attributes strictly for subscriptions and nothing else.

thounyy:
I considered using a struct to enable any attributes in addition to the subscription expiration but it seemed overly complicated for not much. I eventually approached it just as a very simple abstraction of nft_update_attributes(u64) so it is straightforward for devs who want to use it.

If the nft doesn't have u64 attributes, the renew_subscription() function will exit with "error decoding ESDT attributes" because of expires_at() -> get_token_attributes() -> decode_attributes() so there is no risk at this level. This is for developers who want to create nfts exclusively to manage subscriptions, nothing very fancy. It is just an adaptation of the EIP-5643.

I agree that it isn't the most friendly and safe way to add such data in a token but the problem is that the NFT standard on MultiversX is very specific and have multiple dedicated fields for royalties, uris, etc but only one field for arbitrary data (attributes). From my point of view it is better to have single purpose NFTs because of the complexity that adds this single arbitrary field when it comes to handle rich data.

Do you have any suggestion to improve the module @nvergez ?

nvergez:
The attributes are stored in TopEncode, so a developer can store something other than a u64 and not necessarily encounter the `error decoding ESDT attributes`. If what they store fits the top decoding of a `u64`, then your code will interpret it as a `u64`.

But I understand your point. I just want to ensure that a developer doesn't get caught out using this module without knowing its ins and outs, and potentially overwriting the base attributes unintentionally.

andrei-marinica:
Did you mix merge and rebase? I see a lot of our commits re-committed by you, and partially verified. This might mess up our git history. Can you sort them out?

andrei-marinica (approved):

PR #1228, opened 2023-09-27
staking module feedback apply

alyn509:
Security Feedback on Staking Module (from MultiversX framework)

https://docs.google.com/document/d/155ErJ1-rINsDcSu7gdlqcMBwmuiFi3DieVk7VNaicuQ/edit


sasurobert (dismissed):

psorinionut (approved):

jacobleygonie (changes requested):

contracts/modules/src/staking.rs
```diff
@@ -19,12 +21,24 @@ pub trait StakingModule {
...
-        let nr_board_members = user_whitelist.len();
+        for user in user_whitelist {
+            let _ = self.user_whitelist().insert(user);
+        }
+
+        let nr_board_members = self.user_whitelist().len();
         require!(nr_board_members > 0, "No board members");
         require!(
-            slash_quorum <= nr_board_members,
+            nr_board_members <= MAXIMUM_BOARD_MEMBERS,
```
jacobleygonie, line 31: Should we also have this check that `user_whitelist().len() <= MAXIMUM_BOARD_MEMBERS` in in `add_board_member`?

contracts/modules/src/staking.rs
```diff
@@ -102,6 +112,19 @@ pub trait StakingModule {
             .insert(caller);
     }
 
+    #[endpoint(cancelVoteSlashMember)]
+    fn cancel_vote_slash_member(&self, member_to_slash: ManagedAddress) {
+        let caller = self.blockchain().get_caller();
+        require!(
```
jacobleygonie, line 118: I don't think we need this check that board member has stake to cancel a slashing vote.

In fact, we could even argue that it would be better not to have his vote count: so we could let anyone cancel the vote of a board member who is below the staking requirement. In practice maybe this would imply too much code changes, so I would say there is no need to implement this now.

jacobleygonie (approved):

jacobleygonie (approved):

jacobleygonie (approved):

jacobleygonie (approved):

PR #1232, opened 2023-10-09
rust overflow tests

alyn509:


contracts/feature-tests/basic-features/src/small_num_overflow_test_ops.rs
```diff
@@ -0,0 +1,115 @@
+multiversx_sc::imports!();
+
+/// Used for testing overflow on small int types
+#[multiversx_sc::module]
+#[allow(clippy::redundant_clone)]
+pub trait SmallIntOverflow {
+    #[endpoint]
+    #[allow(arithmetic_overflow)]
+    fn usize(&self) -> usize {
```
andrei-marinica, line 9: could we rename them to something like `overflow_usize`, `overflow_*`
just to avoid amiguity in Rust, these are also type names

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #1236, opened 2023-10-10
test-gen logs

mihaicalinluca:


framework/meta/src/cmd/standalone/scen_test_gen/stg_main.rs
```diff
@@ -113,8 +116,10 @@ fn process_file(config: ProcessFileConfig, context: ProcessFileContext) {
         let file_name = format!("{}_{}", context.crate_name, config.suffix);
         context.test_dir.join(file_name)
     };
-    let mut file = File::create(file_path).unwrap();
+
+    let mut file = File::create(file_path.clone()).unwrap();
```
andrei-marinica, line 120: no need to clone, `&file_path` is enough here

framework/meta/src/cmd/standalone/scen_test_gen/stg_main.rs
```diff
@@ -113,8 +116,10 @@ fn process_file(config: ProcessFileConfig, context: ProcessFileContext) {
...
     write!(file, "{new_code}").unwrap();
+    print_new_file(&file_path);
```
andrei-marinica, line 122: Also printed if file exists. You need an additional if.

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #1241, opened 2023-10-12
(dis)allow multiple var args

mihaicalinluca:


framework/base/src/macros.rs
```diff
@@ -202,6 +202,11 @@ macro_rules! only_owner {
     };
 }
 
+#[macro_export]
+macro_rules! allow_multiple_var_args {
```
andrei-marinica, line 206: No need for this.

Our proc-macro gets rid of annotations it processes. This also helps us easier detect cases where attributes are not properly processed, or we missed something.

If it doesn't work without it, it means we missed something.

framework/derive/src/parse/attributes/attr_names.rs
```diff
@@ -25,3 +25,4 @@ pub(super) static ATTR_STORAGE_IS_EMPTY: &str = "storage_is_empty";
 pub(super) static ATTR_STORAGE_CLEAR: &str = "storage_clear";
 pub(super) static ATTR_PROXY: &str = "proxy";
 pub(super) static ATTR_LABEL: &str = "label";
+pub(super) static ATTR_ALLOW_MULTIPLE_VAR_ARGS: &str = "allow_multiple_var_args";
```
andrei-marinica, line 28: Run `cargo fmt`

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #1246, opened 2023-10-23
Hackathon feat

newtmex:
Updates I made during the hackathon.


andrei-marinica:
Looks cool, but we must first make sure that the changes are compatible with the Go implementation.

newtmex:
Alright. Let me know if there's anything I should do.

PR #1249, opened 2023-10-24
sc-meta test CLI

mihaicalinluca:


ovstinga (approved):

framework/meta/src/cli_args/cli_args_standalone.rs
```diff
@@ -76,6 +79,21 @@ pub struct InfoArgs {
...
+pub struct TestArgs {
+    /// Target directory where to generate contract integration tests.
+    /// Will be current directory if not specified.
+    #[arg(short, long, verbatim_doc_comment)]
+    pub path: Option<String>,
+
+    /// This arg can differentiate between types of tests to be ran (go, scenario, all).
+    /// Default value will be "rust" if not specified.
+    #[arg(short, long, default_value = "rust", verbatim_doc_comment)]
+    pub test_type: String,
```
andrei-marinica, line 94: I would use a `--go` argument here, and maybe separately `--scen` for running the scenario tool.

andrei-marinica (changes requested):

framework/meta/src/cli_args/cli_args_standalone.rs
ovstinga: Good idea, it's cleaner this way

andrei-marinica (approved):

PR #1276, opened 2023-11-20
remove wasm renaming in template

alyn509:


BiancaIalangi (approved):

framework/meta/src/cmd/standalone/template/template_adjuster.rs
```diff
@@ -20,7 +19,6 @@ pub struct TemplateAdjuster {
 impl TemplateAdjuster {
     pub fn update_dependencies(&self) {
         self.update_dependencies_root();
-        self.update_dependencies_wasm();
```
andrei-marinica, line 23: We need a check of the template tag (framework version). This can only be removed for v0.45.0 or newer.

andrei-marinica (changes requested):

framework/meta/src/cmd/standalone/template/copy_util.rs
```diff
@@ -1,17 +1,28 @@
...
 /// The function creates all necessary folder structure in the target, no additional preparation required.
-pub fn whitelisted_deep_copy(source_root: &Path, target_root: &Path, whitelist: &[String]) {
+pub fn whitelisted_deep_copy(
+    source_root: &Path,
+    target_root: &Path,
+    whitelist: &[String],
+    args_tag: &str,
+) {
     perform_file_copy(source_root, &PathBuf::new(), target_root, whitelist);
+    create_multiversx_json_file(target_root, args_tag);
```
andrei-marinica, line 25: This is overkill, te file exists in the old versions too, you just need to add its name to the whitelist.

andrei-marinica (changes requested):

andrei-marinica (dismissed):

andrei-marinica (approved):

mihaicalinluca (approved):

PR #1280, opened 2023-11-23
Fill in wasm target dir automatically from workspace, if not explicitly stated

mihaicalinluca:
Speeds up builds with sc-meta.


framework/meta/src/tools/find_workspace.rs
```diff
@@ -0,0 +1,32 @@
...
+        .arg("--message-format=plain")
+        .output()
+        .unwrap()
+        .stdout;
+    let cargo_path = Path::new(std::str::from_utf8(&output).unwrap().trim());
+    print_found_workspace(cargo_path.parent().unwrap());
+    cargo_path.parent().unwrap().to_path_buf()
+}
+
+pub fn print_searching_for_workspace() {
```
andrei-marinica, line 20: This doesn't take long, I don't think we need this message.

framework/meta/src/tools/find_workspace.rs
```diff
@@ -0,0 +1,32 @@
+use colored::Colorize;
+use std::path::{Path, PathBuf};
+
+/// Finds the workspace by searching for the workspace argument into the project's cargo.
+/// Works in debug mode too.
+///
+pub fn find_workspace() -> PathBuf {
```
andrei-marinica, line 7: `find_workspace() -> Option<PathBuf>`

make it recover in case there is no workspace, more versatile this way

I would also split in 2:
find_current_workspace()
find_workspace(path)

(You can give the current path to commands)

Would also be interesting to see if we can write an integration test in meta. It should find the workspace 2 levels up.

framework/meta/src/cmd/contract/sc_config/wasm_build.rs
```diff
@@ -37,6 +38,9 @@ impl ContractVariant {
         }
         if let Some(target_dir_wasm) = &build_args.target_dir_wasm {
             command.args(["--target-dir", target_dir_wasm]);
+        } else {
```
andrei-marinica, line 41: else if let Some(workspace) = find_current_workspace() {...

framework/meta/src/cmd/contract/sc_config/wasm_build.rs
```diff
@@ -37,6 +38,9 @@ impl ContractVariant {
...
+            print_searching_for_workspace();
+            command.args(["--target-dir", find_workspace().to_str().unwrap()]);
```
andrei-marinica, line 43: workspace.join("target")

otherwise it pollutes the actual workspace

framework/meta/src/tools/find_workspace.rs
```diff
@@ -0,0 +1,32 @@
...
+    let output = std::process::Command::new(env!("CARGO"))
+        .arg("locate-project")
+        .arg("--workspace")
+        .arg("--message-format=plain")
+        .output()
+        .unwrap()
+        .stdout;
+    let cargo_path = Path::new(std::str::from_utf8(&output).unwrap().trim());
+    print_found_workspace(cargo_path.parent().unwrap());
```
andrei-marinica, line 16: Don't print here, only where it is used.

framework/meta/src/cmd/contract/sc_config/wasm_build.rs
andrei-marinica: This is where you can print something
Like ...
"Using workspace target directory: {}"

andrei-marinica (changes requested):

framework/meta/src/tools/find_workspace.rs
```diff
@@ -0,0 +1,49 @@
...
+            }
+        }
+    }
+
+    None
+}
+
+pub fn find_workspace(path: &Path) -> Option<PathBuf> {
+    if let Ok(output) = std::process::Command::new(env!("CARGO"))
+        .arg(format!("-C {}", path.display()))
```
andrei-marinica, line 27: `.current_dir(path)`

framework/meta/src/tools/find_workspace.rs
```diff
@@ -0,0 +1,49 @@
+use colored::Colorize;
+use std::path::{Path, PathBuf};
+
+/// Finds the workspace by searching for the workspace argument into the project's cargo.
+/// Works in debug mode too.
+///
+pub fn find_current_workspace() -> Option<PathBuf> {
+    if let Ok(output) = std::process::Command::new(env!("CARGO"))
```
andrei-marinica, line 8: find_workspace(&std::env::current_dir().unwrap())

no duplicate code

framework/meta/src/tools/find_workspace.rs
```diff
@@ -0,0 +1,49 @@
...
+            if let Some(parent) = path.parent() {
+                return Some(parent.to_path_buf());
+            }
+        }
+    }
+
+    None
+}
+
+pub fn print_found_workspace(path: &Path) {
```
andrei-marinica, line 44: "print_workspace_target_dir"

doesn't belong here, it is specific to wasm_build.rs

andrei-marinica (changes requested):

andrei-marinica (approved):

psorinionut (approved):

PR #1289, opened 2023-11-28
Base64 library update

mihaicalinluca:


framework/scenario/src/scenario/model/transaction/tx_response.rs
```diff
@@ -108,7 +110,10 @@ impl TxResponse {
                 return TxResponseStatus::signal_error(&error);
             }
 
-            let error_raw = base64::decode(topics.unwrap().get(1).unwrap()).unwrap();
+            let error_raw = general_purpose::STANDARD
```
andrei-marinica, line 113: would be nice to extract this logic into a base64_decode function and place it somewhere in the crate root

andrei-marinica (changes requested):
There is also a place where there is still a deprecated base64::encode, it shows up as warning. Maybe also extract that to a function.

PR #1290, opened 2023-11-28
Bitflags library update

mihaicalinluca:


andrei-marinica (changes requested):

framework/wasm-adapter/src/api/blockchain_api_node.rs
```diff
@@ -356,10 +356,9 @@ impl BlockchainApiImpl for VmApiImpl {
...
     ) -> multiversx_sc::types::EsdtLocalRoleFlags {
-        unsafe {
-            multiversx_sc::types::EsdtLocalRoleFlags::from_bits_unchecked(getESDTLocalRoles(
-                token_id_handle,
-            ) as u64)
-        }
+        multiversx_sc::types::EsdtLocalRoleFlags::from_bits(unsafe {
+            getESDTLocalRoles(token_id_handle)
+        } as u64)
+        .unwrap()
```
andrei-marinica, line 362: Unwrap not allowed here, since it can panic.
We need to find a better solution.
Specifically in this case I think error can be ignored

andrei-marinica (changes requested):

framework/wasm-adapter/src/api/blockchain_api_node.rs
andrei-marinica: in case of error, call
`error_hook::signal_error(message)`
on else

andrei-marinica (approved):

BiancaIalangi (approved):

PR #1391, opened 2024-01-19
Install mx-scenario-go

andrei-marinica:


mihaicalinluca (approved):

andreibancioiu (approved):
In the future, all constants from the GitHub workflow should be handled differently.

PR #1394, opened 2024-01-22
check when FailAllocator is triggered

BiancaIalangi:


framework/meta/src/tools/wasm_extractor.rs
```diff
@@ -0,0 +1,90 @@
...
+pub fn is_memory_allocation(wasm_data: Vec<u8>) -> Result<bool, BinaryReaderError> {
+    let parser = Parser::new(0);
+
+    for payload in parser.parse_all(&wasm_data).flatten() {
+        if let Payload::DataSection(data_section) = payload {
+            for data_fragment in data_section {
+                let cleaned_data_fragment: Vec<u8> = data_fragment?
+                    .data
+                    .iter()
+                    .filter(|&&b| b < 128)
```
andrei-marinica, line 76: What does this condition mean?

BiancaIalangi: There were cases where fragments taken from wasm payload contained a non-UTF-8 sequence. Because of this, it could no longer be converted into a String. This condition detects and removes invalid bytes that can't be converted to a character.

andrei-marinica: I think it's safer to not convert to String, and search for the message as bytes

framework/meta/src/tools/wasm_extractor.rs
```diff
@@ -0,0 +1,90 @@
...
+            "error occured while extracting memory allocation information from .wasm: file not found",
+        );
+
+        let allocator_trigger = is_memory_allocation(wasm_data)
+            .expect("error occured while extracting memory allocation information from .wasm ");
+
+        if allocator_trigger {
+            println!(
+                "{}",
+                "FailAllocator triggered: memory allocation forbidden"
```
andrei-marinica, line 39: "FailAllocator used while memory allocation is accessible in code. Contract may fail unexpectedly when memory allocation is attempted".

andrei-marinica (approved):

mihaicalinluca (approved):

PR #1395, opened 2024-01-22
CI install via sc-meta

andrei-marinica:


andreibancioiu (approved):
All hardcoded versions (Rust, sc-scenario, mxpy) should be handled differently in the future.

PR #1426, opened 2024-02-12
New TxProxies

BiancaIalangi:


framework/meta/src/cli_args/cli_args_contract.rs
```diff
@@ -65,7 +65,13 @@ pub enum ContractCliAction {
...
         about = "Generates a snippets project, based on the contract ABI."
     )]
-    GenerateSnippets(GenerateSnippetsArgs),
+    GenerateSnippets(GenerateOverwriteArg),
+
+    #[command(
+        name = "proxy",
+        about = "Generates a proxy, based on the contract ABI."
+    )]
+    GenerateProxies(GenerateOverwriteArg),
```
andrei-marinica, line 74: Don't reuse the same arg type for both snippets and proxies. Even if, by coincidence they currently contain the same fields, they are likely to eveolve in different directions in the future.

framework/meta/src/cmd/contract/generate_proxy/proxy_crate_gen.rs
```diff
@@ -0,0 +1,15 @@
+use std::fs::File;
+
+#[must_use]
+pub(crate) fn create_file(proxies_file_name: &str, overwrite: bool) -> File {
+    let file = format!("../{proxies_file_name}");
+
+    if overwrite {
+        File::create(&file).unwrap()
```
andrei-marinica, line 8: instead of .unwrap(), better `.expect("could not write proxy file")`

andrei-marinica (changes requested):

andrei-marinica (approved):

mihaicalinluca (approved):

PR #1440, opened 2024-02-20
remove maybe uninit

alyn509:


framework/base/src/types/managed/wrapped/managed_vec.rs
```diff
@@ -147,12 +153,14 @@ where
...
-        let mut result_uninit = core::mem::MaybeUninit::<T::Ref<'_>>::uninit_array();
+        let mut result_uninit =
+            unsafe { MaybeUninit::<[MaybeUninit<T::Ref<'_>>; N]>::uninit().assume_init() };
+
         for (index, value) in self.iter().enumerate() {
             result_uninit[index].write(value);
         }
 
-        let result = unsafe { core::mem::MaybeUninit::array_assume_init(result_uninit) };
+        let result = unsafe { transmute_copy(&ManuallyDrop::new(result_uninit)) };
```
andrei-marinica, line 163: I think there is a more efficient way, that avoids the copy, with just "transmute", which changes the pointer type.

andrei-marinica (changes requested):

framework/base/src/types/managed/wrapped/managed_vec.rs
alyn509: use of unstable library feature 'core_intrinsics': intrinsics are unlikely to ever be stabilized, instead they should be used through stabilized interfaces in the rest of the standard library
add `#![feature(core_intrinsics)]` to the crate attributes to enable

there might not be another way to workaround this except if we find another way to get the reference without `MaybeUninit`

alyn509: the whole implementation of those rely on those intrinsics

we can get rid of 
```
#![feature(maybe_uninit_uninit_array)]
#![feature(maybe_uninit_array_assume_init)]
```

by adding 

`#![feature(core_intrinsics)]`

andrei-marinica (approved):

BiancaIalangi (approved):

PR #1442, opened 2024-02-21
Design for TxProxy

BiancaIalangi:


contracts/examples/adder/proxies_struct_interactor_main.rs
```diff
@@ -1,60 +1,90 @@
...
-impl<A> TxProxy<A>
-where
-    A: multiversx_sc::api::VMApi + 'static,
-{	fn init<
-		Arg0: multiversx_sc::codec::CodecInto<BigUint<A>>,
+impl<Env: TxEnv + multiversx_sc::api::CallTypeApi> TxProxyMethods<Env> {
+	pub fn init<
+		Arg0: multiversx_sc::codec::CodecInto<BigUint<Env>>,
 	>(
 		&mut self,
```
andrei-marinica, line 20: Make it take `self` directly. Thid object wil be single use.

contracts/examples/adder/proxies_struct_interactor_main.rs
```diff
@@ -1,60 +1,90 @@
...
-		let ___opt_address___ = multiversx_sc::extract_opt_address!(self);
-		let mut ___contract_deploy___ = multiversx_sc::constructors_proxy!(___opt_address___);
-		___contract_deploy___.push_endpoint_arg(&initial_value);
-		___contract_deploy___
+	) -> multiversx_sc::types::Tx<Env,
+        (),
+        (),
+        (),
+        (),
+        FunctionCall<<Env as multiversx_sc::types::TxEnv>::Api>,
```
andrei-marinica, line 27: `init` produces a `DeployCall` here. 

Also, I'm pretty sure that in this case we can write `<Env::Api>` directly, hygiene should not be an issue, and it looks easier to read.

contracts/examples/adder/proxies_struct_interactor_main.rs
```diff
@@ -1,60 +1,90 @@
...
+        (),
+    > {
+		Tx::new_with_env(self.env.clone())
```
andrei-marinica, line 30: Since we're having `self` instead of `&mut self`, this clone can happily disappear.

framework/derive/src/generate/proxy_gen.rs
```diff
@@ -298,6 +298,24 @@ pub fn proxy_trait(contract: &ContractTrait) -> proc_macro2::TokenStream {
     }
 }
 
+pub fn tx_proxy_trait() -> proc_macro2::TokenStream {
+    quote! {
+        pub trait TxProxyTrait<Env>{
```
andrei-marinica, line 303: There is only one `TxProxyTrait` and one `TxProxyMethods`, they don't need to be generated for each contract. You can place them directly somewhere in types/interaction, in framework base.

The derive crate remains untouched in thie PR.

andrei-marinica (changes requested):

PR #1443, opened 2024-02-22
upgrade constructor impl for unified syntax + fmt

mihaicalinluca:


BiancaIalangi (approved):

alyn509 (approved):

framework/meta/src/cmd/contract/sc_config/contract_variant_validate.rs
```diff
@@ -16,21 +16,32 @@ fn check_single_constructor(contract_variant: &ContractVariant) -> Result<(), St
             Err("Missing constructor. Add a method annotated with `#[init]`.".to_string())
         },
         1 => Ok(()),
-        _ => Err("More than one contrctructor present. Exactly one method annotated with `#[init]` is required.".to_string()),
```
andrei-marinica, line 19: I think we can reorganize this to look better/simpler. How about
```
match (init_constructor.len(), upgrade_constructors.len()) {
    (0, 0) => nope
    (1, 0,) | (0, 1) | (1, 1) => Ok(()),
    (_, _) => nope
}
```

andrei-marinica (changes requested):
Code looks great, but we're still missing an ABI test in abi-tester

andrei-marinica (approved):

PR #1460, opened 2024-03-07
Add type name rust abi

BiancaIalangi:


framework/meta/src/abi_json/type_abi_json.rs
```diff
@@ -68,10 +68,13 @@ impl From<&TypeDescription> for TypeDescriptionJson {
 }
 
 impl TypeDescriptionJson {
-    pub fn to_type_description(&self, name: &str) -> TypeDescription {
+    pub fn to_type_description(&self, name_abi: &str, name_rust: &str) -> TypeDescription {
```
andrei-marinica, line 71: `names: TypeNames`, cleaner

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #1474, opened 2024-03-12
remove never type and exhaustive patterns unstable features

mihaicalinluca:


data/codec/src/codec_err_handler.rs
```diff
@@ -34,12 +34,15 @@ impl DecodeErrorHandler for DefaultErrorHandler {
     }
 }
 
+#[derive(Debug)]
+pub enum Never {}
```
andrei-marinica, line 38: Turns out is already exists, we should reuse the existing one: https://doc.rust-lang.org/core/convert/enum.Infallible.html

data/codec/src/equivalent/codec_convert.rs
```diff
@@ -9,9 +9,14 @@ where
     Medium: Default + TopDecodeMultiInput + TopEncodeMultiOutput,
 {
     let mut medium: Medium = Default::default();
-    let Ok(()) = from.multi_encode_or_handle_err(&mut medium, PanicErrorHandler);
-    let Ok(result) = To::multi_decode_or_handle_err(&mut medium, PanicErrorHandler);
-    result
+    match from.multi_encode_or_handle_err(&mut medium, PanicErrorHandler) {
+        Ok(()) => (),
+        Err(err) => panic!("panic occured: {:#?}", err),
```
andrei-marinica, line 14: `unreachable!()` instead of panic

Also: we should provide a function or trait with one method to extract the data out of Result<T, Infallible>, and we should reuse that. There is too much duplicated code.

Something like ".unwrap_infallible()".

".unwrap()" is also fine, but maybe having a separate name makes it easier for us to revert, once the exhaustive patterns get stabilized.

andrei-marinica: Somebody else thought about this. Import, or just copy code
https://docs.rs/unwrap-infallible/latest/src/unwrap_infallible/lib.rs.html#1-99

framework/base/src/types/io/sc_error_static.rs
```diff
@@ -54,8 +57,8 @@ impl From<DecodeError> for StaticSCError {
     }
 }
 
-impl From<!> for StaticSCError {
-    fn from(_: !) -> Self {
+impl From<Never> for StaticSCError {
```
andrei-marinica, line 60: (Can probably be deleted with no consequences)

andrei-marinica (changes requested):

BiancaIalangi (approved):

PR #1481, opened 2024-03-17
chore: add typo checker in CI

tabVersion:
as title


andrei-marinica:
Great initiative, love it!
Our company's policy is to have all commits signed, so I'd be grateful if you could redo your commits signed.
To fix the failing tests I think some files also need to be renamed accordingly.

tabVersion:
> Great initiative, love it! Our company's policy is to have all commits signed, so I'd be grateful if you could redo your commits signed. To fix the failing tests I think some files also need to be renamed accordingly.

I've redo the commits and signed. Thanks for your comments

andrei-marinica:
Added here: https://github.com/multiversx/mx-sdk-rs/pull/2106

PR #1484, opened 2024-03-19
New config in `sc-config.toml`: `proxy-paths`

BiancaIalangi:
It is now possible to configure in `sc-config.toml` where to save the generated proxy.


mihaicalinluca (dismissed):

framework/meta/src/cmd/contract/sc_config/sc_config_serde.rs
```diff
@@ -62,6 +62,9 @@ pub struct ContractVariantSerde {
 #[serde(deny_unknown_fields)]
 pub struct MultiContractGeneralSettingsSerde {
     pub main: Option<String>,
+
+    #[serde(rename = "proxy-paths")]
```
andrei-marinica, line 66: no need for the Option
only `#[serde(default)]`

contracts/examples/adder/sc-config.toml
```diff
@@ -0,0 +1,4 @@
+[settings]
+proxy-paths = ["src/adder_proxy.rs"]
+
+[contracts.adder]
```
andrei-marinica, line 4: Let's try to make this part not mandatory.

To explain: if this file is missing, a default config is created. Let's make this also work when this fie exists, but contracts are not configured.

andrei-marinica (changes requested):

andrei-marinica (approved):

mihaicalinluca (approved):

andrei-marinica:
Re-created here: https://github.com/multiversx/mx-sdk-rs/pull/1489

PR #1494, opened 2024-03-22
Replacing direct transfers with unified syntax

BiancaIalangi:


contracts/modules/src/esdt.rs
```diff
@@ -63,9 +63,9 @@ pub trait EsdtModule {
             ManagedAsyncCallResult::Err(_) => {
                 // return payment to initial caller
                 let initial_caller = self.blockchain().get_owner_address();
-                let egld_returned = self.call_value().egld_value();
-                if *egld_returned > 0u32 {
-                    self.send().direct_egld(&initial_caller, &egld_returned);
+                let egld_returned = self.call_value().egld_value().to_u64().unwrap();
```
andrei-marinica, line 66: We shouldn't convert to u64. Might easily not fit. Just dropping the conversion, and it should still work.

contracts/feature-tests/composability/forwarder/src/esdt.rs
```diff
@@ -73,13 +73,11 @@ pub trait ForwarderEsdtModule: storage::ForwarderStorageModule {
...
-            &all_token_payments,
-            self.blockchain().get_gas_left(),
-            &ManagedBuffer::new(),
-            &ManagedArgBuffer::new(),
-        );
+        self.tx()
+            .with_gas_limit(self.blockchain().get_gas_left())
+            .to(&to)
+            .multi_esdt(all_token_payments)
+            .transfer_execute();
```
andrei-marinica, line 80: .transfer() should work just fine.

framework/base/src/types/interaction/tx_call_te.rs
```diff
@@ -49,4 +49,12 @@ where
     pub fn transfer(self) {
         self.transfer_execute_with_gas(0)
     }
+
+    pub fn transfer_non_zero(self) {
```
andrei-marinica, line 53: Add some docs with what it does "Transfers funds, amount is greater than zero. Does nothing otherwise"

andrei-marinica (changes requested):

PR #1500, opened 2024-03-23
BlockchainWrapper - caller & sc address handle optimization

andrei-marinica:


andrei-marinica:
PR is profoundly unsafe and not worth the optimization.

PR #1515, opened 2024-03-27
proxy pause sc migration to new proxy and unified syntax

mihaicalinluca:


contracts/examples/proxy-pause/src/pause_sc_proxy.rs
```diff
@@ -0,0 +1,99 @@
+// Code generated by the multiversx-sc proxy generator. DO NOT EDIT.
```
andrei-marinica, line 1: You changed it by hand, delete the prelude. also the allows.

contracts/examples/proxy-pause/src/pause_sc_proxy.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[rustfmt::skip]
+impl<Env, From, Gas> PausableProxyMethods<Env, From, (), Gas>
+where
+    Env: TxEnv,
+    Env::Api: VMApi,
+    From: TxFrom<Env>,
+    Gas: TxGas<Env>,
+{
+    pub fn init(
```
andrei-marinica, line 46: init is not part of this interface

andrei-marinica (changes requested):

BiancaIalangi (approved):

PR #1516, opened 2024-03-27
encode error from basic features migration to new proxy and unified syntax

mihaicalinluca:
…yntax


contracts/feature-tests/basic-features/src/codec_err_test.rs
```diff
@@ -62,24 +48,24 @@ pub trait CodecErrorTest {
...
     /// It just covers contract call serialization errors.
     #[endpoint]
     fn codec_err_contract_call(&self) {
         let _ = self
-            .encode_err_proxy()
-            .contract(ManagedAddress::zero())
+            .tx()
+            .to(&ManagedAddress::zero())
+            .typed(encode_error_proxy::EncodeErrorProxy)
             .encode_error_method(CodecErrorTestType);
```
andrei-marinica, line 69: This just verifies the encoding part. Adding the full generated proxy is overkill. We should just
```
let _ = self.tx().raw_call().argument(CodecErrorTestType);
```

This proxy was just easier in the past, can be dropped.

andrei-marinica (changes requested):

andrei-marinica (approved):

contracts/feature-tests/basic-features/src/codec_err_test.rs
```diff
@@ -62,24 +48,17 @@ pub trait CodecErrorTest {
...
-
-    #[proxy]
-    fn encode_err_proxy(&self) -> encode_err_proxy::Proxy<Self::Api>;
-
     /// Never actually calls any deploy/upgrade, so it is appropriate in this contract.
     /// It just covers contract init serialization errors.
     #[endpoint]
     fn codec_err_contract_init(&self) {
-        let _ = self.encode_err_proxy().init(CodecErrorTestType);
+        let _ = self.tx().raw_call().argument(&CodecErrorTestType);
```
andrei-marinica, line 55: .raw_deploy()

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #1518, opened 2024-03-28
Fix proxy

BiancaIalangi:


framework/meta/src/cmd/contract/generate_proxy/proxy_generator.rs
```diff
@@ -0,0 +1,496 @@
...
+                field.name,
+                self.adjust_type_name_with_api(&field.field_type.rust)
+            )
+            .unwrap();
+        }
+
+        writeln!(file, "    }},").unwrap();
+    }
+
+    fn clean_paths(&self, proxy_crate: &str, rust_type: &str) -> String {
```
andrei-marinica, line 441: To see if this function works as expected, let's write some unit tests for it.

andrei-marinica (changes requested):

andrei-marinica (approved):

mihaicalinluca (approved):

andrei-marinica (approved):

PR #1522, opened 2024-03-29
contracts migration to new proxies and unified syntax

mihaicalinluca:


BiancaIalangi (approved):

contracts/feature-tests/composability/builtin-func-features/src/builtin_func_proxy.rs
```diff
@@ -1,10 +1,68 @@
-multiversx_sc::imports!();
+// Code generated by the multiversx-sc proxy generator. DO NOT EDIT.
 
-#[multiversx_sc::derive::proxy]
-pub trait UserBuiltin {
-    #[endpoint(SetUserName)]
-    fn set_user_name(&self, name: &ManagedBuffer);
+////////////////////////////////////////////////////
```
andrei-marinica, line 3: This is an example where there should be no prefix, but I'll fix it myself later today.

andrei-marinica (approved):

PR #1538, opened 2024-04-05
multi exec interact unified syntax

mihaicalinluca:


andrei-marinica:
Fixed in https://github.com/multiversx/mx-sdk-rs/pull/1541

PR #1543, opened 2024-04-08
codec - universal_decode_number optimziation

andrei-marinica:


andrei-marinica:
Fixed here: https://github.com/multiversx/mx-sdk-rs/pull/1597

PR #1559, opened 2024-04-12
set state builder impl and tests

mihaicalinluca:


contracts/feature-tests/scenario-tester/tests/st_blackbox_test.rs
```diff
@@ -1,36 +1,65 @@
...
+
     let st_contract = ContractInfo::<scenario_tester::Proxy<StaticApi>>::new("sc:scenario-tester");
 
     world.start_trace();
 
-    world.set_state_step(
-        SetStateStep::new()
-            .put_account(owner_address, Account::new().nonce(1))
-            .new_address(owner_address, 1, "sc:scenario-tester"),
+    world
```
andrei-marinica, line 34: Let's have a test with just these 2: a set stater followed by check state. This will help us in the long run.

We can test with and without .commit() at the end.

andrei-marinica (dismissed):

andrei-marinica (approved):

PR #1592, opened 2024-04-26
set state via AnnotatedValue

andrei-marinica:


framework/scenario/src/facade/world_tx/scenario_set_state/scenario_set_account.rs
```diff
@@ -30,123 +39,144 @@ impl SetStateBuilderItem for AccountItem {
...
-        balance_expr: V,
-        opt_attributes_expr: Option<T>,
-        royalties_expr: N,
-        creator_expr: Option<T>,
-        hash_expr: Option<T>,
-        uris_expr: Vec<T>,
+        token_id: K,
+        nonce: N,
+        balance: V,
+        attributes: T,
```
andrei-marinica, line 110: We need separate types here for each.

framework/scenario/src/facade/world_tx/scenario_set_state/scenario_set_account.rs
```diff
@@ -30,123 +39,144 @@ impl SetStateBuilderItem for AccountItem {
...
+        royalties: N,
+        creator: T,
```
andrei-marinica, line 112: Should be another type, this one an address

mihaicalinluca (approved):

BiancaIalangi (approved):

PR #1595, opened 2024-04-29
codec - small number decode optimization (part 2)

andrei-marinica:


andrei-marinica:
Fixed here: https://github.com/multiversx/mx-sdk-rs/pull/1597

PR #1623, opened 2024-05-10
TypeAbiFrom tuples

BiancaIalangi:


andrei-marinica:
We realized it cannot be implemented like this for tuples.
Implemented for MultiValueN here https://github.com/multiversx/mx-sdk-rs/pull/1625

PR #1679, opened 2024-06-10
Generate custome proxy

BiancaIalangi:


framework/meta-lib/src/contract/sc_config/contract_variant_builder.rs
```diff
@@ -315,7 +388,7 @@ impl ScConfig {
                 wasm_crate_name,
                 abi: original_abi.clone(),
             }],
-            proxy_configs: Vec::new(),
+            proxy_configs: HashMap::new(),
```
andrei-marinica, line 391: Proxies still form a Vec, there is no key

framework/meta-lib/src/contract/sc_config/sc_config_model.rs
```diff
@@ -18,7 +20,7 @@ pub const SC_CONFIG_FILE_NAMES: &[&str] = &["sc-config.toml", "multicontract.tom
 pub struct ScConfig {
     pub default_contract_config_name: String,
     pub contracts: Vec<ContractVariant>,
-    pub proxy_configs: Vec<ProxyConfigSerde>,
+    pub proxy_configs: HashMap<ProxyConfigSerde, ContractVariant>,
```
andrei-marinica, line 23: It's still a Vec

framework/meta-lib/src/contract/sc_config/contract_variant_builder.rs
```diff
@@ -253,40 +254,112 @@ fn validate_contract_variants(contracts: &[ContractVariant]) {
...
+    set_main_contract_flag(&mut contracts, &config.settings.main);
+    validate_contract_variants(&contracts);
+
+    contracts
+}
+
+fn process_proxy_contracts(
+    config: &ScConfigSerde,
+    original_abi: &ContractAbi,
+) -> HashMap<ProxyConfigSerde, ContractVariant> {
```
andrei-marinica, line 287: Should be a Vec

andrei-marinica: The config should not be the key here, it should be part of the proxy config
You need to add the ContractVariant to the ProxyConfig

andrei-marinica (changes requested):

andrei-marinica (dismissed):

mihaicalinluca (approved):

andrei-marinica (approved):

PR #1694, opened 2024-07-02
upgrade fn gen in sc-meta snippets

mihaicalinluca:


framework/meta-lib/src/contract/generate_snippets/snippet_template_gen.rs
```diff
@@ -55,6 +55,15 @@ async fn main() {{
     // all contracts have a deploy snippet
     writeln!(file, r#"        "deploy" => interact.deploy().await,"#).unwrap();
 
+    for upgrade_endpoint in &abi.upgrade_constructors {
+        writeln!(
+            file,
+            r#"        "{}" => interact.{}().await,"#,
+            upgrade_endpoint.name, upgrade_endpoint.rust_method_name
```
andrei-marinica, line 62: upgrade_endpoint.name is always "upgrade"
make it a constant somewhere (or use the one that already exists)

andrei-marinica: In fact, there shouldn't be a `name` field there, it's redundant, since only one value is allowed

andrei-marinica, line 62: We need pre-processing protection for upgrade and init

andrei-marinica (changes requested):

BiancaIalangi (dismissed):

framework/derive/src/parse/attributes/payable_attr.rs
```diff
@@ -36,6 +36,26 @@ fn extract_token_identifier(attr: &syn::Attribute) -> Option<String> {
                         !clean.is_empty(),
                         "ticker can not be empty. attribute needs 1 string argument: Replace with #[payable(\"*\")] or #[payable(\"EGLD\")"
                     );
+
+                    assert!(
```
andrei-marinica, line 40: Would be nice to extract these checks into a separate function

framework/derive/src/parse/attributes/util.rs
```diff
@@ -103,11 +103,31 @@ pub(super) fn attr_one_string_arg(attr: &syn::Attribute) -> String {
             let mut iter = list.tokens.into_iter();
             let arg_token_tree = match iter.next() {
                 Some(proc_macro2::TokenTree::Literal(literal)) => {
+                    let clean = literal
```
andrei-marinica, line 106: Would be nice to extract these checks into a separate function

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #1701, opened 2024-07-05
sync call readonly missing impl

mihaicalinluca:


andrei-marinica:
Fixed in https://github.com/multiversx/mx-sdk-rs/pull/1702

PR #1704, opened 2024-07-10
example for nightly version shift

BiancaIalangi:


andreiblt1304 (approved):

andrei-marinica:
Doesn't look like we want this in the repo.

PR #1705, opened 2024-07-11
Endpoints module rename

BiancaIalangi:


mihaicalinluca (approved):

contracts/benchmarks/large-storage/wasm/src/lib.rs
```diff
@@ -14,7 +14,7 @@
 multiversx_sc_wasm_adapter::allocator!();
 multiversx_sc_wasm_adapter::panic_handler!();
 
-multiversx_sc_wasm_adapter::endpoints! {
+multiversx_sc_wasm_adapter::__wasm__endpoints__! {
```
andrei-marinica, line 17: This is a macro, not the endpoints module. Please do not rename this one.

andrei-marinica (changes requested):

PR #1709, opened 2024-07-16
added support for custom structs and small upgrade and import fixes

mihaicalinluca:


framework/meta-lib/src/contract/generate_snippets/snippet_sc_functions_gen.rs
```diff
@@ -83,6 +83,39 @@ fn write_deploy_method_impl(file: &mut File, init_abi: &EndpointAbi, name: &Stri
...
+    let proxy_name = format!("{}Proxy", name);
+
+    writeln!(
+        file,
+        r#"        let response = self
+            .interactor
+            .tx()
+            .to(self.state.current_address())
+            .from(&self.wallet_address)
+            .gas(NumExpr("{DEFAULT_GAS}"))
```
andrei-marinica, line 98: Maybe we should just use u64, it's more straightforward, less mandos baggage

andrei-marinica (approved):

BiancaIalangi (approved):

PR #1724, opened 2024-08-02
Add events for PauseModule

CostinCarabas:


contracts/modules/src/pause.rs
```diff
@@ -48,6 +48,12 @@ pub trait PauseModule {
         require!(self.not_paused(), "Contract is paused");
     }
 
+    #[event("pause_contract")]
```
dragos-rebegea, line 51: i think the event is usually cammelCase

mihaicalinluca: in the sdk there are examples with both cases, do you prefer camelCase for this one?

BiancaIalangi (approved):

sasurobert (approved):

PR #1728, opened 2024-08-12
chain simulator - integration with interactor

BiancaIalangi:


sdk/core/src/gateway/gateway_proxy.rs
```diff
@@ -1,17 +1,23 @@
...
 #[derive(Clone, Debug)]
 pub struct GatewayProxy {
     pub(crate) proxy_url: String,
     pub(crate) client: Client,
+    pub chain_simulator: bool,
 }
 
 impl GatewayProxy {
     pub fn new(proxy_url: String) -> Self {
+        let chain_simulator = proxy_url == CHAIN_SIMULATOR_GATEWAY;
```
andrei-marinica, line 15: We should allow callers to set their own URL, who knows, maybe at some point it will be running on a different port.

We should pass the URL from the interactor

andrei-marinica (changes requested):

mihaicalinluca (dismissed):

contracts/feature-tests/composability/interact/src/comp_interact_config.rs
```diff
@@ -11,7 +11,8 @@ const CONFIG_FILE: &str = "config.toml";
 /// Multisig Interact configuration
 #[derive(Debug, Deserialize)]
 pub struct Config {
-    gateway: String,
+    gateway_uri: String,
+    chain_type: String,
```
andrei-marinica, line 15: Let's make this an enum

andrei-marinica (changes requested):

mihaicalinluca (approved):

andrei-marinica (approved):

mihaicalinluca (approved):

PR #1737, opened 2024-08-19
Keystore signing done

JustEatAnApple:


sdk/core/src/wallet.rs
```diff
@@ -131,4 +165,70 @@ impl Wallet {
...
         self.priv_key.sign(tx_bytes)
     }
+
+    pub fn validate_keystore_password(path: &str) -> Result<DecryptionParams, WalletError> {
+        println!(
+            "Insert password. Press 'Ctrl-D' (Linux / MacOS) or 'Ctrl-Z' (Windows) when done."
+        );
+        let mut password = String::new();
+        io::stdin().read_to_string(&mut password).unwrap();
+        password = password.trim().to_string();
```
andrei-marinica, line 175: This prompt should be extracted somewhere else. Let's have the password given as parameter.

andrei-marinica (changes requested):

sdk/core/src/wallet.rs
JustEatAnApple: Done

andrei-marinica (approved):

BiancaIalangi (approved):

PR #1753, opened 2024-08-28
Modified the adder interactor example to work with keystore-signing

JustEatAnApple:
We modified the function from_keystore_secret to get the password as an argument instead of requesting it from the terminal.

### DIFF
--- a/contracts/examples/adder/interact/src/basic_interact.rs
+++ b/contracts/examples/adder/interact/src/basic_interact.rs
@@ -209,13 +209,43 @@ impl AdderInteract {
         println!("sum: {sum}");
     }
 
+    async fn fail_upgrade(&mut self, new_value: u32) {
+        self.interactor
+            .tx()
+            .from(&self.wallet_address)
+            .to(self.state.current_adder_address())
+            .gas(6_000_000)
+            .typed(adder_proxy::AdderProxy)
+            .upgrade(BigUint::from(new_value))
+            .code_metadata(CodeMetadata::UPGRADEABLE)
+            .code(ADDER_CODE_PATH)
+            .with_result(ExpectError(4, "upgrade is allowed only for owner"))
+            .prepare_async()
+            .run()
+            .await;
+
+        let sum = self
+            .interactor
+            .query()
+            .to(self.state.current_adder_address())
+            .typed(adder_proxy::AdderProxy)
+            .sum()
+            .returns(ReturnsResultUnmanaged)
+            .prepare_async()
+            .run()
+            .await;
+        assert_ne!(sum, RustBigUint::from(new_value));
+
+        println!("Upgrade failed: upgrade is allowed only for owner");
+    }
+
     async fn upgrade(&mut self, new_value: u32) {
         let response = self
             .interactor
             .tx()
-            .from(&self.wallet_address)
+            .from(&self.adder_owner_address)
             .to(self.state.current_adder_address())
-            .gas(3_000_000)
+            .gas(6_000_000)
             .typed(adder_proxy::AdderProxy)
             .upgrade(BigUint::from(new_value))
             .code_metadata(CodeMetadata::UPGRADEABLE)
@@ -243,7 +273,18 @@ impl AdderInteract {
 
 #[tokio::test]
 #[ignore = "run on demand"]
-async fn test() {
+async fn fail_upgrade_test() {
+    let mut basic_interact = AdderInteract::init().await;
+
+    basic_interact.deploy().await;
+    basic_interact.add(1u32).await;
+
+    basic_interact.fail_upgrade(7u32).await;
+}
+
+#[tokio::test]
+#[ignore = "run on demand"]
+async fn upgrade_test() {
     let mut basic_interact = AdderInteract::init().await;
 
     basic_interact.deploy().await;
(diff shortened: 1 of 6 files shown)


BiancaIalangi (approved):

contracts/examples/adder/interact/src/basic_interact.rs
```diff
@@ -242,4 +250,4 @@ async fn test() {
     basic_interact.add(1u32).await;
 
     basic_interact.upgrade(7u32).await;
```
mihaicalinluca, line 252: Let's add a parameter of type ExpectError and have 2 transactions, one that fails as seen here (from non owner address) and one that's successful (from owner address) so that the test passes if run. Also upgrade and deploy need more than 3 mil gas in this case

mihaicalinluca (changes requested):

contracts/examples/adder/interact/src/basic_interact.rs
JustEatAnApple: Upgraded the code a little, should work as intended now

contracts/examples/adder/interact/src/basic_interact.rs
```diff
@@ -201,13 +209,43 @@ impl AdderInteract {
         println!("sum: {sum}");
     }
 
+    async fn fail_upgrade(&mut self, new_value: u32) {
```
mihaicalinluca, line 212: Instead of duplicating the code, just add an ExpectError parameter and a from parameter to the upgrade function and call it twice in the same test. Catch the error using .returns

mihaicalinluca (changes requested):

contracts/examples/adder/interact/src/basic_interact.rs
```diff
@@ -7,7 +7,7 @@ use basic_interact_config::Config;
 use basic_interact_state::State;
 use clap::Parser;
 
-use multiversx_sc_snippets::imports::*;
+use multiversx_sc_snippets::{imports::*, sdk::data::keystore::InsertPassword};
```
andrei-marinica, line 10: Let's add `sdk::data::keystore::InsertPassword` to the imports, so it's easier for developers

contracts/examples/adder/interact/src/basic_interact.rs
```diff
@@ -65,7 +68,15 @@ impl AdderInteract {
...
-        let wallet_address = interactor.register_wallet(test_wallets::mike());
+        // PASSWORD: "alice"
+        // InsertPassword::Plaintext("alice".to_string()) || InsertPassword::StandardInput
+        let wallet_address = interactor.register_wallet(
+            Wallet::from_keystore_secret(
+                "alice.json",
+                InsertPassword::Plaintext("alice".to_string()),
+            )
+            .unwrap(),
+        );
```
andrei-marinica, line 79: Awsome!

contracts/examples/adder/interact/src/basic_interact.rs
```diff
@@ -201,45 +212,87 @@ impl AdderInteract {
...
+                println!("response: {response:?}");
+            },
+            None => {
+                self.interactor
+                    .tx()
+                    .from(sender)
+                    .to(self.state.current_adder_address())
+                    .gas(6_000_000)
+                    .typed(adder_proxy::AdderProxy)
+                    .upgrade(BigUint::from(new_value))
```
andrei-marinica, line 247: converison not needed, `.upgrade(new_value)` should work

contracts/examples/adder/interact/src/basic_interact.rs
```diff
@@ -201,45 +212,87 @@ impl AdderInteract {
...
-            .returns(ReturnsResultUnmanaged)
-            .prepare_async()
-            .run()
-            .await;
-        assert_eq!(sum, RustBigUint::from(new_value));
+    async fn upgrade(
+        &mut self,
+        new_value: u32,
+        sender: &Bech32Address,
+        expected_result: Option<(u64, &str)>,
```
andrei-marinica, line 219: Interactors aren't really tests, so we don't really expect errors. We should probably test the interactor framework somewhere else.

That being said, it works for now, can be left like this. Much better would be to have a result handler that can a return a Result<T, Error>, but this is a tricky task that's been postponed a lot.

andrei-marinica (changes requested):

contracts/examples/adder/interact/src/basic_interact.rs
JustEatAnApple: Done

contracts/examples/adder/interact/src/basic_interact.rs
JustEatAnApple: Done

contracts/examples/adder/interact/src/basic_interact.rs
JustEatAnApple: For now this is the cleanest solution that we thought of. In time, if the Result<T, Error> becomes available we will upgrade it

andrei-marinica (approved):

BiancaIalangi (approved):

mihaicalinluca (approved):

PR #1761, opened 2024-09-03
returnsLogs result handler impl and tests

mihaicalinluca:


framework/snippets/src/network_response.rs
```diff
@@ -58,16 +59,72 @@ fn process_out(tx: &TransactionOnNetwork) -> Vec<Vec<u8>> {
...
             if event.identifier == "writeLog" {
                 if let Some(data) = &event.data {
-                    let decoded_data = String::from_utf8(base64_decode(data)).unwrap();
-
-                    if decoded_data.starts_with('@') {
-                        let out = decode_scr_data_or_panic(decoded_data.as_str());
-                        return Some(out);
+                    match data {
+                        Data::String(string) => {
+                            let decoded_data = String::from_utf8(base64_decode(string)).unwrap();
```
andrei-marinica, line 108: Duplicate code. Easier to just convert to vec first:
```
let data_vec = match data {
    Data::String(string) =>vec![string],
    Data::Vec(vec) => vec,
};

for data_member in vec.iter() { ...
```

Better still, create a method in Data that outputs a vec

andrei-marinica (changes requested):

andrei-marinica (approved):

BiancaIalangi (approved):

PR #1767, opened 2024-09-06
added support for SCResults without a data field, avoid unnecessary p…

mihaicalinluca:
…anic

### DIFF
--- a/framework/snippets/src/network_response.rs
+++ b/framework/snippets/src/network_response.rs
@@ -52,12 +52,14 @@ fn process_success(tx: &TransactionOnNetwork) -> TxResponse {
 fn process_out(tx: &TransactionOnNetwork) -> Vec<Vec<u8>> {
     let out_scr = tx.smart_contract_results.iter().find(is_out_scr);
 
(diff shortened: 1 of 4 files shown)


framework/snippets/src/network_response.rs
```diff
@@ -52,10 +52,12 @@ fn process_success(tx: &TransactionOnNetwork) -> TxResponse {
 fn process_out(tx: &TransactionOnNetwork) -> Vec<Vec<u8>> {
     let out_scr = tx.smart_contract_results.iter().find(is_out_scr);
 
-    if let Some(out_scr) = out_scr {
-        decode_scr_data_or_panic(&out_scr.data)
-    } else {
-        process_out_from_log(tx).unwrap_or_default()
+    match out_scr {
```
andrei-marinica, line 55: if let is the standard for handling Option, no need to use `match`

andrei-marinica (dismissed):

mihaicalinluca:
Issue solved in a different PR.
Tests re-added in #1843.

PR #1771, opened 2024-09-12
Detect write in storage ops in view methods

BiancaIalangi:


framework/meta-lib/src/tools/wasm_extractor.rs
```diff
@@ -1,31 +1,40 @@
...
 
 use super::report_creator::{ReportCreator, WITHOUT_MESSAGE, WITH_MESSAGE};
 
+type CallGraph = HashMap<usize, HashSet<usize>>;
+
 const PANIC_WITH_MESSAGE: &[u8; 16] = b"panic occurred: ";
 const PANIC_WITHOUT_MESSAGE: &[u8; 14] = b"panic occurred";
 const ERROR_FAIL_ALLOCATOR: &[u8; 27] = b"memory allocation forbidden";
 const MEMORY_GROW_OPCODE: u8 = 0x40;
+const WRITE_OP: [&str; 1] = ["mBufferStorageStore"];
```
andrei-marinica, line 21: 1. &[&str] = &[...]; - static slice is easier to write and cleaner
2. there are more write hooks: storageStore, int64storageStore, bigIntStorageStoreUnsigned, smallIntStorageStoreUnsigned, smallIntStorageStoreSigned
3. we should probably extend this to contract calls & transfers, they shouldn't be performed in a view, I think; event logs are also questionable

framework/meta-lib/src/tools/wasm_extractor.rs
```diff
@@ -44,50 +54,136 @@ fn populate_wasm_info(
...
-        has_format: true,
+        imports: wasm_info.imports,
+        ei_check: wasm_info.ei_check,
+        memory_grow_flag: wasm_info.memory_grow_flag,
+        call_graph: wasm_info.call_graph,
         report,
     })
 }
 
+fn detect_write_operations_in_views(
```
andrei-marinica, line 106: It would be nice to also provide a unit test. It shouldn't be too difficult, you can embed the adder as a WAT.

andrei-marinica (changes requested):

mihaicalinluca (approved):

PR #1775, opened 2024-09-20
wasm extractor - fix report parameters

BiancaIalangi:


andrei-marinica (approved):

framework/meta-lib/src/tools/wasm_extractor.rs
```diff
@@ -49,25 +49,35 @@ fn populate_wasm_info(
...
+                ei_check |= is_ei_valid(imports.clone(), check_ei);
             },
             Payload::DataSection(data_section) => {
-                allocator_trigger = is_fail_allocator_triggered(data_section.clone());
-                if is_panic_with_message_triggered(data_section.clone()) {
-                    has_panic = WITH_MESSAGE;
-                } else if is_panic_without_message_triggered(data_section) {
-                    has_panic = WITHOUT_MESSAGE;
+                allocator_trigger |= is_fail_allocator_triggered(data_section.clone());
+                match has_panic {
```
andrei-marinica, line 63: This match belongs to a method in `PanicMessage`. Something like `max_severity`. We should also have a separate function that gives the PanicMessage for the current section.

We should also rename it to something like PanicReport or PanicStatus, since it is a flag, rather than just a message.

mihaicalinluca (approved):

andrei-marinica (approved):

PR #1837, opened 2024-10-28
Allow for cross-compilation MacOS -> Linux any arch

gfusee:
I added a static-tls features in the snippets and sdk crates. This feature, when enabled, use a rust implementation of tls, instead of relying on the system's OpenSSL installation.

The goal is to allow developers to cross-compile their projects, especially when the host is a mac arm64 to a linux target (any arch).


gfusee:
will recreate it on rc/v0.54

PR #1869, opened 2024-11-20
set state implementation for chain simulator

mihaicalinluca:


contracts/examples/adder/interactor/src/basic_interactor.rs
```diff
@@ -57,7 +57,7 @@ impl AdderInteract {
         let adder_owner_address = interactor.register_wallet(test_wallets::heidi()).await;
         let wallet_address = interactor.register_wallet(test_wallets::ivan()).await;
 
-        interactor.generate_blocks_until_epoch(1).await.unwrap();
+        let _ = interactor.generate_blocks(30u64).await;
```
andrei-marinica, line 60: I think we still need the unwrap, to stop execution in case of error

contracts/examples/adder/interactor/tests/basic_interactor_cs_test.rs
```diff
@@ -33,3 +34,30 @@ async fn simulator_upgrade_test() {
...
+
+#[tokio::test]
+#[cfg_attr(not(feature = "chain-simulator-tests"), ignore)]
+async fn set_state_cs_test() {
+    let account_address = test_wallets::mike();
+
+    let real_chain_interact = AdderInteract::new(Config::load_config()).await;
+    let simulator_interact = AdderInteract::new(Config::chain_simulator_config()).await;
+
+    let account = real_chain_interact
```
andrei-marinica, line 46: We should save the account somewhere.
These are separate actions.
The most common scenario will be:
1. We download the state once, save it to some file, or in code. Mandos SetState is already implemented, to be decided if we need something else.
2. We read from file, set the state to chain simulator and run our test.

It is also possible to simulate starting from real chain, but this is not reliable as a test, since the mainnet state might be compromised. So we should keep this function as well, but not as an integration test.

framework/snippets/src/interactor/interactor_chain_simulator.rs
```diff
@@ -53,4 +54,14 @@ where
             ))
             .await
     }
+
+    pub async fn set_state(&self, account: Vec<SetStateAccount>) -> Result<String, Error> {
```
andrei-marinica, line 58: `accounts`

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #1908, opened 2024-12-23
[DO NO MERGE] EgldOrEsdtTokenIdentifier encode as EGLD-000000 instead of EGLD

andrei-marinica:
It has backwards compatibility issues.

Kept for future reference.


andrei-marinica:
This is extremely dangerous for storage. While values can be still safely be decoded, appropriate keys will not be found.

To change the encoding, a new type is needed.

PR #1913, opened 2025-01-08
esdt transfer raw tests

dorin-iancu:


contracts/feature-tests/composability/builtin-func-features/src/esdt_features.rs
```diff
@@ -0,0 +1,64 @@
+multiversx_sc::imports!();
+multiversx_sc::derive_imports!();
+
+static ESDT_TRANSFER_FUNC_NAME: &[u8] = b"ESDTTransfer";
+
+const GAS_LIMIT_ESDT_TRANSFER: u64 = 50_0000;
```
CostinCarabas, line 6: I think the value proposed is too low. Also, you can have a better format.

CostinCarabas (approved):

andreiblt1304 (approved):

contracts/feature-tests/composability/builtin-func-features/src/esdt_features.rs
CostinCarabas: Then the format should have been "500_000"

dorin-iancu: You're right, fixed

andreiblt1304 (approved):

CostinCarabas (approved):

PR #1923, opened 2025-01-16
naming fix that was preventing set state in chain sim

mihaicalinluca:


BiancaIalangi (dismissed):

sdk/core/src/gateway/gateway_chain_simulator_set_state.rs
```diff
@@ -55,8 +55,8 @@ impl From<Account> for SetStateAccount {
 }
 
 impl SetStateAccount {
-    pub fn with_keys(mut self, keys: HashMap<String, String>) -> Self {
```
andrei-marinica, line 58: Backwards compatibility! Need to keep both!

andrei-marinica (changes requested):

andrei-marinica (dismissed):

andrei-marinica (approved):

BiancaIalangi (approved):

PR #1938, opened 2025-01-23
fix: update readme

aneiosi:
I updated the readme to remove the reference to the deprecated vscode extension


andrei-marinica:
Hi! We need all commits to be signed to accept PRs.

PR #1940, opened 2025-01-28
colored dependency - bump version

BiancaIalangi:


BiancaIalangi:
`colored = 3.0.0` fails on nightly rust

PR #1967, opened 2025-02-13
role retrieving from SystemSC storage

mihaicalinluca:


CostinCarabas (approved):

contracts/feature-tests/basic-features/src/special_roles_from_system_account.rs
```diff
@@ -0,0 +1,22 @@
+use multiversx_sc::storage::StorageKey;
+
+multiversx_sc::imports!();
+
+/// Exposes ways to retrieve special roles of a specific token on-chain
+#[multiversx_sc::module]
+pub trait RetrieveSpecialRoles {
+    #[endpoint]
+    fn token_has_transfer_role(&self, token_id: TokenIdentifier) -> bool {
+        let key = ManagedBuffer::new_from_bytes(b"ELRONDtransferesdt");
```
andrei-marinica, line 10: Can use the new `token_has_transfer_role`, isn't it?

framework/base/src/contract_base/wrappers/storage_raw_wrapper.rs
```diff
@@ -86,4 +87,18 @@ where
         let key: StorageKey<A> = storage_key.into();
         storage_set(key.as_ref(), value);
     }
+
+    pub fn token_has_transfer_role(&self, token_identifier: TokenIdentifier<A>) -> bool {
```
andrei-marinica, line 91: I would move it to the BlockchaiWrapper, there are more ESDT-related methods there.

andrei-marinica (changes requested):

contracts/feature-tests/basic-features/src/special_roles_from_system_account.rs
mihaicalinluca: I thought maybe it would be useful to leave both examples, but I can only leave the one using the wrapper

andrei-marinica (approved):

PR #1974, opened 2025-02-21
path improvements for better compatibility

mihaicalinluca:


framework/meta/src/cmd/all.rs
```diff
@@ -7,15 +7,15 @@ use std::{path::Path, process::Command};
 
 pub fn call_all_meta(args: &AllArgs) {
     let path = if let Some(some_path) = &args.path {
-        some_path.as_str()
+        Path::new(some_path)
     } else {
-        "./"
+        Path::new("./")
```
andrei-marinica, line 12: does this work on windows?

andrei-marinica (approved):

framework/meta/src/cmd/all.rs
mihaicalinluca: yes, Path knows how to deal with it

andrei-marinica (approved):

BiancaIalangi (approved):

PR #1978, opened 2025-02-24
Pretty log error reporting

alyn509:


framework/scenario/src/scenario/run_vm/tx_output_check.rs
```diff
@@ -51,20 +70,33 @@ pub fn check_tx_output(tx_id: &str, tx_expect: &TxExpect, tx_result: &TxResult)
...
+                        ),
                     )
                 }
             }
         },
     }
 }
 
+fn format_result_message(message: &CheckValue<BytesValue>) -> String {
+    let mut formatted_message = message.to_string();
```
andrei-marinica, line 90: not like that

use trim() to get rid of the quotes
.strip_prefix("str") to get rid of the prefix

andrei-marinica (changes requested):

framework/scenario/src/scenario/run_vm/tx_output_check.rs
alyn509: funny enough, `trim()` works only for whitespaces, but not quotes. 

I sanitized a bit the code tho

andrei-marinica (approved):

mihaicalinluca (approved):

BiancaIalangi (approved):

andreiblt1304 (approved):

PR #2015, opened 2025-03-13
validate args before execution in sc-meta - extra checks for sc-meta args

mihaicalinluca:


framework/meta/src/cli/cli_args_validate.rs
```diff
@@ -0,0 +1,130 @@
...
+            if !VERSIONS.iter().any(|v| v.to_string() == *override_target_v) {
+                user_error(&format!("Invalid requested version: {}", override_target_v));
+            }
+        }
+    }
+}
+
+// helpers
+pub(crate) fn validate_contract_name(name: &str) -> bool {
+    let valid_name_regex = Regex::new(r"^[a-zA-Z_][a-zA-Z0-9_]*$").unwrap();
```
andrei-marinica, line 116: Importing a big regex crate might be a bit overkill, and might slow down builds.

Should be just a few lines of code.

if name.is_empty() return false;
if !name[0].is_ascii_alphabetic() return false;
for c in name { if !name[0].is_ascii_alphanumeric() && c != '-' && c != '_' return false; }
return true

BiancaIalangi (approved):

andrei-marinica (approved):

PR #2017, opened 2025-03-17
getTokenProperties interaction

mihaicalinluca:


framework/base/src/types/interaction/result_handlers.rs
```diff
@@ -10,6 +10,7 @@ mod returns_raw_result;
 mod returns_result;
 mod returns_result_as;
 mod returns_result_unmanaged;
+mod token_properties_result;
```
andrei-marinica, line 13: Not here, it's not a result handler.

Either in types/io, or in types/interaction/system_proxy, where it is also used

framework/base/src/types/interaction/result_handlers/token_properties_result.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub can_transfer_nft_create_role: bool,
+    pub nft_create_stopped: bool,
+    pub num_wiped: usize,
+}
+
+impl TokenPropertiesResult {
+    fn fetch_struct_field(&mut self, input: &[u8]) {
+        let is_true = |value: &[u8]| value == b"true";
+
+        if let Some(value) = input.strip_prefix(b"NumDecimals-") {
```
andrei-marinica, line 28: Put all prefixes in a const slice and iterate over them (except the last one). This will lead to smaller contracts and less duplicated code.

framework/scenario/tests/token_properties_result_test.rs
```diff
@@ -0,0 +1,48 @@
+use multiversx_sc::{codec::TopDecodeMulti, types::TokenPropertiesResult};
+
+fn hex_decode(hex: &[u8]) -> Option<Vec<u8>> {
```
andrei-marinica, line 3: We already have a call data decode facility, it's called HexCallDataDeserializer. Maybe it's worth picking it up and dusting it off.

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #2036, opened 2025-04-16
Opcode detect

BiancaIalangi:


framework/meta-lib/src/tools/wasm_extractor.rs
```diff
@@ -40,30 +40,33 @@ impl WasmInfo {
...
+    fn create_call_graph(&mut self, body: FunctionBody) -> Result<(), WasmError> {
         let mut instructions_reader = body
             .get_operators_reader()
             .expect("Failed to get operators reader");
 
         let mut call_functions = HashSet::new();
         while let Ok(op) = instructions_reader.read() {
+            if !is_whitelisted(&op) {
+                let op_str = format!("{:?}", op);
+                return Err(WasmError::ForbiddenOpcode(op_str));
```
andrei-marinica, line 67: I think it would be smarter to collect them all, instead of exiting on the first one. Easier to investigate, in case they exist.

framework/meta-lib/src/tools/wasm_extractor.rs
```diff
@@ -262,3 +271,137 @@ fn is_mem_grow(code_section: &FunctionBody) -> bool {
 
     false
 }
+
+fn is_whitelisted(op: &Operator) -> bool {
```
andrei-marinica, line 275: Even though we are not auto-generating it directly right now, it would be nice to be in a separate file.

andrei-marinica (changes requested):

framework/meta-lib/src/tools/wasm_extractor/extractor_tests.rs
```diff
@@ -427,70 +434,77 @@ pub mod tests {
...
 
     #[test]
     fn test_empty_with_fail_allocator() {
         if let Ok(content) = Parser::new().parse_bytes(None, EMPTY_WITH_FAIL_ALLOCATOR.as_bytes()) {
-            let wasm_info =
-                populate_wasm_info(Path::new(""), content.to_vec(), false, &None, Vec::new())
-                    .expect("Unable to parse WASM content.");
-            assert!(!wasm_info.memory_grow_flag);
-            assert!(wasm_info.report.has_allocator);
+            let wasm_info = populate_wasm_info(
```
andrei-marinica, line 451: populate_wasm_info has a lot of arguments, sounds like a WasmInfo object with a builder pattern might make it clearer

andrei-marinica (approved):

PR #2051, opened 2025-05-15
vm - VMHooksHandler converted to struct

andrei-marinica:
Inheritance should be avoided.
Nesting is preferred.

VMHooksHandler used to be a trait, connected to more traits, and based on trait VMHooksSource. A bit complicated.

The nested version looks like this: VMHooksHandler is a struct, with VM hooks implementations spread into several files with `impl` fields. The source (VMHooksSource) is now a field, instead of a supertrait.


framework/scenario/src/api/impl_vh/debug_api.rs
```diff
@@ -23,7 +23,7 @@ impl VMHooksApiBackend for DebugApiBackend {
     {
         let instance = ContractDebugStack::static_peek();
         let tx_context_ref = instance.tx_context_ref.clone();
-        let handler = TxContextVMHooksHandler::new(tx_context_ref, ContractDebugInstanceState);
+        let handler = TxVMHooksContext::new(tx_context_ref, ContractDebugInstanceState);
```
andrei-marinica, line 26: context

BiancaIalangi (approved):

mihaicalinluca (approved):

PR #2054, opened 2025-05-19
wasm target - autoinstall

BiancaIalangi:


framework/meta-lib/src/contract/sc_config/wasm_build.rs
```diff
@@ -54,6 +79,19 @@ impl ContractVariant {
         command
     }
 
+    fn compose_rustup_command(&self, arguments: Vec<&str>) -> Command {
```
andrei-marinica, line 82: rustup_command

framework/meta-lib/src/contract/sc_config/wasm_build.rs
```diff
@@ -54,6 +79,19 @@ impl ContractVariant {
...
+        let rustup = env::var_os("RUSTUP").unwrap_or_else(|| OsString::from("rustup"));
+
+        let mut command = Command::new(rustup);
+        command.arg("target");
+
+        for argument in arguments {
```
andrei-marinica, line 88: Let these be set by the caller

andrei-marinica (approved):

PR #2090, opened 2025-06-16
Barnard: ESDT token type API

BiancaIalangi:


framework/base/src/contract_base/wrappers/blockchain_wrapper.rs
```diff
@@ -157,6 +157,27 @@ where
...
 
+    #[cfg(feature = "barnard")]
+    pub fn get_esdt_token_type(
+        &self,
+        address: &ManagedAddress<A>,
+        token_id: &TokenIdentifier<A>,
+        nonce: u64,
+    ) -> EsdtTokenType {
+        unsafe {
+            let result: ManagedBuffer<A> = ManagedBuffer::new_uninit();
```
andrei-marinica: It's not a managed buffer, it's a big int in the VM.

You can also use a temp handle.

andrei-marinica (changes requested):

PR #2115, opened 2025-07-09
Governance interactor

BiancaIalangi:


contracts/feature-tests/payable-features/interactor/src/payable_interactor.rs
```diff
@@ -48,6 +48,9 @@ impl PayableInteract {
             .await
             .use_chain_simulator(config.use_chain_simulator());
 
+        interactor
+            .set_current_dir_from_workspace("contracts/feature-tests/payable-features/interactor");
```
andrei-marinica, line 52: Do we still need it?

BiancaIalangi: Yes, otherwise the mxsc file won't be found

andrei-marinica (approved):

PR #2122, opened 2025-07-18
TypeAbiFrom impl for u128

BiancaIalangi:


contracts/feature-tests/payable-features/interactor/src/payable_interactor.rs
```diff
@@ -91,7 +91,7 @@ impl PayableInteract {
             .typed(payable_features_proxy::PayableFeaturesProxy)
             .payable_all_transfers()
             .payment(payment)
-            .returns(ReturnsResult)
+            .returns(ReturnsRawResult)
```
andrei-marinica: This is suspicious. It has nothing to do with the current PR. We need to investigate further.

andrei-marinica (approved):

PR #2131, opened 2025-08-01
deprecate `call_and_exit`

alyn509:


andreiblt1304 (dismissed):

andreiblt1304 (approved):

framework/base/src/types/interaction/tx_exec/tx_exec_async.rs
```diff
@@ -141,6 +141,11 @@ where
         )
     }
 
+    /// Backwards compatibility only.
+    #[deprecated(
+        since = "0.59.2",
+        note = "Backwards compatibility only, does nothing. Just delete. Use `async_call_and_exit` to launch asynchronous calls or `sync_call` to launch synchronous calls."
```
BiancaIalangi: The message is too verbose. Try: "Please use method `async_call_and_exit` instead."

PR #2142, opened 2025-08-19
add extra data to `get_token_properties` query

BiancaIalangi:


andreiblt1304 (approved):

chain/core/src/types/flags/esdt_token_type.rs
```diff
@@ -38,6 +38,18 @@ impl EsdtTokenType {
             EsdtTokenType::Invalid => 255,
         }
     }
+
+    pub fn collection_from_string(token_type: String) -> Self {
```
andrei-marinica: this crate is no_std
better avoid String
I see &str is sufficient

andrei-marinica (changes requested):

andreiblt1304 (approved):

andrei-marinica (approved):

PR #2147, opened 2025-09-04
removed legacy contract call framework

andrei-marinica:


andrei-marinica:
Handled in https://github.com/multiversx/mx-sdk-rs/pull/2163 and https://github.com/multiversx/mx-sdk-rs/pull/2161

PR #2168, opened 2025-09-25
storage v2 - context and key

BiancaIalangi:


framework/base/src/storage_v2/key.rs
```diff
@@ -0,0 +1,410 @@
...
+use crate::{
+    api::{
+        ErrorApi, ManagedTypeApi, StorageReadApi, StorageReadApiImpl, StorageWriteApi,
+        StorageWriteApiImpl,
+    },
+    storage::StorageKey,
+    types::{ManagedBuffer, ManagedType},
+};
+
+#[derive(Default, Clone)]
```
andrei-marinica: No need for wrapped object, use StorageKey directly.

I only used dynamic key in the prototype to distinguish from the const keys. I was thinking of StorageKey from the get go.

PR #2354, opened 2026-04-21
sc-meta source pack

andrei-marinica:


andrei-marinica:
Merged as part of github.com/multiversx/mx-sdk-rs/pull/2355

